Issue #48, opened 2023-05-09
Need support for static methods


migueldeicaza:
FileAccess.open is static.

migueldeicaza:
Done

PR #49, opened 2023-05-13
Fix duplication in static keyword for method generation

alicerunsonfedora:
Currently, the generator ends up creating methods with the static keyword repeated, causing a build failure:

```swift
public static static func () {}
```

This PR attempts to fix this, as well as address a build error regarding gi.object_method_bind_ptrcall in cases of static methods. I don't know how well this solution works, since I couldn't really test any static methods available to me. (I was hoping this would generate the utility functions).

Edit: This should help contribute to #48 .

### DIFF
--- a/Generator/Generator/ClassGen.swift
+++ b/Generator/Generator/ClassGen.swift
@@ -223,6 +223,8 @@ func generateMethods (_ p: Printer,
             }
             if instanceOrStatic == "" {
                 finalp = "final "
+            } else if instanceOrStatic == "static" {
+                finalp = " "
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thank you for the fix, I had a similar fix for the double static static, but I think you are correct about the `nil`.

When I redid this code, I was going by the C# generated code, but I now realize that the C# code actually is using a different mechanism to call into Godot, so the removal of the nil instance is wrong.

I am merging this change manually, due to the other file change that I am not sure what it is

PR #53, opened 2023-05-16
Add load and print family of utilities in GD

alicerunsonfedora:
This PR adds new `load`, `print`, and `pushError`/`pushWarning` methods to the global scope GD class.

The load function works similarly to that of GD.Load in the C# variant and is a means of being consistent with the GDScript preload function.

Three new print methods are also available, which function similarly to Swift's traditional print function. These methods also handle converting the items accordingly to ensure they get represented properly, as well as converting the resulting string to the Variant type that the normal GD.print methods require.


migueldeicaza:
Thank you for sending this patch!

PR #70, opened 2023-05-28
Update build scripts to add arm64 to Mac builds

alicerunsonfedora:
The xcframework now contains a universal framework on the macOS side for arm64 and x86_64. The two are stitched together using lipo.

### DIFF
--- a/scripts/make-swiftgodot-framework
+++ b/scripts/make-swiftgodot-framework
@@ -22,9 +22,15 @@ esac
 
 rm -rf $2
 bdMac=$1/Build/Products/Debug/
+bdMacArm="$(echo "$1" | rev | cut -c2- | rev)_arm/Build/Products/Debug/"
 bdiOS=$1/Build/Products/Debug-iphoneos/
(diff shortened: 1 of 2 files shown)


alicerunsonfedora:
Hmm, I might need to update this to disable the pushing to GitHub and all that, since I don't think it should be needed when wanting to build it locally (as in my case).

migueldeicaza:
Thanks for the patch - I am struggling with this line, and need to sort it out:

```
xcodebuild -scheme SwiftGodot -destination platform=generic/macOS,arch=arm64
```

It produces:
```
xcodebuild: error: Unable to find a device matching the provided destination specifier:
		{ platform:generic/macOS, arch:arm64 }
```

My Xcode shows that macOS builds are built-in.

I'll set up an M1 to do the builds and see if that can sort it out.

alicerunsonfedora:
That is particularly odd, though I suspect that it only shows up for Apple Silicon Macs.

alicerunsonfedora:
I'm closing this PR to have this in a dedicated brance

PR #74, opened 2023-05-29
Add Apple Silicion support in XCFramework

alicerunsonfedora:
Supersedes #70.


migueldeicaza:
Thank you!  

I still do not know why my Intel Mac shows absolutely no presence or indication of an ARM SDK, but my ARM one has both.  And Google does not show much.

The ARM machine feels faster than the Intel one, it does the 3 builds in 3 minutes.

PR #78, opened 2023-05-31
Fixing the SwiftPM example

seivan:
~~So it can be generated with `swift package generate-xcodeproj`~~
Edit: Nevermind that, it's deprecated. 
Still fixes issues with semicolons and matching closing tags.


migueldeicaza:
Thanks for your patch!

Issue #83, opened 2023-06-06
Can't initialize color?


nvanfleet:
It looks like it's not possible to initialize color from Swift. The value just ends up as completely black or sometimes contains infinite values. Looking at the code below:

```
        let color = Color(r: 1, g: 1, b: 1, a: 1)
        let variant = Variant(color)
        let color2 = Color(variant)
        GD.print("color ", color, color.toHtml(withAlpha: true))
        GD.print("color2 ", color2, color2?.toHtml(withAlpha: true))
        let color3 = Color(r: 0.5, g: 0.5, b: 0.5, a: 0.5)
        GD.print("test color 3", color3, color3.toHtml(withAlpha: true))

        // A bit of a sanity test but I was assuming it was between 0 and 1
        let color4 = Color(r: 255, g: 255, b: 255, a: 255)
        GD.print("test color 4", color4, color4.toHtml(withAlpha: true))
        
        // This will actually work
        let colorCode = Color(code: "#959846")
        GD.print("colorCode ", colorCode, colorCode.toHtml(withAlpha: true))
        let variant2 = Variant(colorCode)
        let variantColorCode = Color(variant2)
        GD.print("colorCode from variant", variantColorCode, variantColorCode?.toHtml(withAlpha: true))
```

```
color  Color(red: 0.0, green: 0.0, blue: 0.0, alpha: 0.0) 00000000
color2  Optional(SwiftGodot.Color(red: 0.0, green: 0.0, blue: 0.0, alpha: 0.0)) Optional("00000000")
test color 3 Color(red: 0.0, green: 0.0, blue: 0.0, alpha: 0.0) 00000000
test color 4 Color(red: 0.0, green: 0.0, blue: 0.0, alpha: inf) 000000ff
colorCode  Color(red: 0.58431375, green: 0.59607846, blue: 0.27450982, alpha: 1.0) 959846ff
colorCode from variant Optional(SwiftGodot.Color(red: 0.58431375, green: 0.59607846, blue: 0.27450982, alpha: 1.0)) Optional("959846ff")
```

I noticed that running this a few times would sometimes randomly make the red color either inf or -inf but I am not sure why since the input into the color init was not changing.

nvanfleet:
I created some more tests which I ran in Godot and oddly it seems like it's only Color that has issues at least according to these simple tests.

```
private func testTypes() {
        GD.print("Testing")
        
        let object = SwiftGodot.Object()
        
[...]

PR #84, opened 2023-06-07
Replace instances of Float with Double

nvanfleet:
I noticed that when initializing color you could create a Color value of Color(r: 1, g: 1, b: 1, a: 1) and somehow only get Color(r: 0, g: 0, b: 0, a: 0) with some random numbers sometimes in place of the zeros. With some testing I realized that using Double for Color instead solved the issue. So it seems like the cpp expects doubles and not floats.

Note: this actually seems to cause issues with Vector. So it's possible that it's not as cut and dry and replacing all instances of Float with Double. 

https://github.com/migueldeicaza/SwiftGodot/issues/83


nvanfleet:
After some more testing even the most minimal change here to fix Color breaks things like Vector. So I'm not completely certain how to split the hairs here to make this work.

Issue #86, opened 2023-06-08
Macros: Exporting methods as Godot methods, without having to do custom registration or being limited to Variant parameters.


migueldeicaza:
Currently when we register methods, we have to register method that take variants, we declare what the types of those variant arguments are, and then we return a variant.   Then user code has to extract the variant, and return a variant.

With macros, we could instead generate all of this glue on behalf of the user, so they can use regular Swift code and Swift signatures without having to deal with any of those problems.

migueldeicaza:
Current progress:

```
@Godot
class Demo {
      @Export
      public func computeGodot (x: String, y: Int) -> Double {
        return 1.0
      }
  
      @Export
      public func computeOptional (x: String, y: Double) -> Double? {
        return nil
      }
  
      public func compute (x: String, y: Int) -> Double {
        return 1.0
      }
      public var property: Int { 1 }
      
      @Export
      public var godotProperty: Int { 1 }
  }
```

Generates these artifacts:

```
static var initClass: Void = {
    let className = StringName("Demo")
    let classInfo = ClassInfo<Demo> (name: className)
    let prop_0 = PropInfo (propertyType: .float, propertyName: "", className: className, hint: .none, hintStr: "", usage: .propertyUsageDefault)
    let prop_1 = PropInfo (propertyType: .string, propertyName: "x", className: className, hint: .none, hintStr: "", usage: .propertyUsageDefault)
    let prop_2 = PropInfo (propertyType: .int, propertyName: "y", className: className, hint: .none, hintStr: "", usage: .propertyUsageDefault)
    let computeGodotArgs = [
        prop_1,
        prop_2,
    ]
    classInfo.registerMethod(name: "funcName", flags: .default, returnValue: prop_0, arguments: computeGodotArgs, function: _mproxy_computeGodot)    let prop_3 = PropInfo (propertyType: .float, propertyName: "y", className: className, hint: .none, hintStr: "", usage: .propertyUsageDefault)
    let computeOptionalArgs = [
        prop_1,
        prop_3,
    ]
    classInfo.registerMethod(name: "funcName", flags: .default, returnValue: prop_0, arguments: computeOptionalArgs, function: _mproxy_computeOptional)
}

func _mproxy_computeGodot (args: [Variant]) -> Variant? {
    let result = computeGodot (String (args [0]!), Int (args [1]!))
    guard let result else { return nil }
    return Variant (result)
}

[...]

PR #89, opened 2023-06-08
Add convenience registration methods, fix build scripts

alicerunsonfedora:
This PR adds some new helper methods I've written to improve the general process of registering certain types of data fields:

- `registerCheckbox` registers a toggleable boolean value, rendered as a checkbox in the editor.
- `registerEnum` registers an enum that can be expressed as an integer. The signature calls for conformance to `Nameable` which would allow the names to show up correctly in the editor.
- `registerInt` registers a range picker for an integer value.
- `registerFilePicker` registers a file picker to pick from a series of files. For macOS 10.15, this accepts a string array of the file extensions. For macOS 11.0+, this accepts an array of uniform type idenfitiers (UTType). In turn, this also includes some UTType instance for various Godot-specific files such as `tscn` and `tres`.
- `registerTextField` registers a text field with a string that can be manipulated.
- `registerTextView` is similar to `registerTextField`, but supporting multiline strings.

For writing setter methods, one can call the `ClassInfo.withCheckedProperty` method to take care of typical work for ensuring values exist:

```swift
func setProperty(args: [Variant]) -> Variant? {
    withCheckedProperty(named: "magic", args: args) { argument in
        // Do something here...
    }
}
```

---

This PR also makes some minor tweaks to the build scripts to build the universal binaries _only when run on Apple Silicon devices_. Because GitHub Actions agents don't have Apple Silicon runners yet, building the framework fails.

(I have CI scripts on my end that build the framework instead of calling `swift build`).

### DIFF
--- a/Sources/SwiftGodot/Extensions/ClassInfo+ConvenienceProperties.swift
+++ b/Sources/SwiftGodot/Extensions/ClassInfo+ConvenienceProperties.swift
@@ -37,6 +37,55 @@ extension ClassInfo {
     /// A type alias referencing a registerable int enum.
     public typealias RegisteredIntEnum = CaseIterable & Nameable & RawRepresentable<Int>
 
+    /// Performs an operation on an argument that originates from a setter method.
+    ///
+    /// This can be used inside setter method to set a property in-class with a guaranteed argument value.
+    ///
+    /// ```swift
+    /// func setBubbleCount(args: [Variant]) -> Variant? {
+    ///     withCheckedProperty(named: "bubbles", in: args) { argument in
+    ///         self.bubbles = Int(argument) ?? 0
+    ///     }
+    /// }
+    /// ```
+    ///
+    /// - Parameter name: The name of the property that is being set.
+    /// - Parameter arguments: The list of arguments that were passed from the setter.
+    /// - Parameter action: A closure that accepts a valid argument.
+    public static func withCheckedProperty(named name: String,
+                                           in arguments: [Variant],
+                                           perform action: (Variant) -> Void) -> Variant? {
+        guard let arg = arguments.first else {
+            GD.pushError("Expected argument for \(name), but got nil instead.")
+            return nil
+        }
+        action(arg)
+        return nil
+    }
+
+    /// Registers a checkbox toggle in the editor.
+    /// - Parameter name: The name of the property that will appear in the editor.
+    /// - Parameter prefix: The prefix to apply to the property name. Defaults to the class's name if not provided.
+    /// - Parameter getter: The getter method the editor will call to get the property.
+    /// - Parameter setter: The setter method the editor will call to set the property.
+    public func registerCheckbox(named name: String,
+                                 prefix: String? = nil,
+                                 getter: @escaping ClassInfoFunction,
+                                 setter: @escaping ClassInfoFunction) {
+        let registeredPrefix = prefix ?? "\(T.self)"
+        let property = PropInfo(propertyType: .bool,
+                                propertyName: StringName("\(registeredPrefix)_\(name)"),
+                                className: StringName("\(T.self)"),
+                                hint: .flags,
+                                hintStr: "",
+                                usage: .propertyUsageDefault)
+        registerSetter(prefix: registeredPrefix, name: name, property: property, setter: setter)
+        registerGetter(prefix: registeredPrefix, name: name, property: property, getter: getter)
+        registerProperty(property,
+                         getter: StringName("\(registeredPrefix)_get_\(name)"),
+                         setter: StringName("\(registeredPrefix)_set_\(name)"))
+    }
+
     /// Registers an enumeration in the editor.
     /// - Parameter name: The name of the property that will appear in the editor.
     /// - Parameter enumType: The enumeration type that will be selected in the editor.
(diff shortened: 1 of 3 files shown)


alicerunsonfedora:
@migueldeicaza I can imagine that these helper methods may help with writing the macros to auto-register these properties.

migueldeicaza:
Thanks for this patch!

I am guessing that some of these methods are helpers that are invoked from the class initialization process?

My only reservation is that this might be duplicating functionality of a macro-based approach.   I have not finished watching the Macro videos, nor started experimenting, but it feels like we could achieve most of this with macros (I filed a few bugs with more detailed information of what I have in mind).

My big caveat with Macros right now is if I can fully automate the class generation

alicerunsonfedora:
> I am guessing that some of these methods are helpers that are invoked from the class initialization process?

Yes, these would be called in the class initialization process:

```swift
extension AnthroCharacterBody2D {
    static var initClass: Void = {
        let classInfo = ClassInfo<AnthroCharacterBody2D>(name: "AnthroCharacterBody2D")
        classInfo.registerEnum(named: "character",
                               for: Character.self,
                               prefix: "player",
                               getter: AnthroCharacterBody2D.getCharacter,
                               setter: AnthroCharacterBody2D.setCharacter)

        classInfo.addPropertyGroup(name: "Movement", prefix: "movement_")
        classInfo.registerInt(named: "acceleration",
                              range: 1 ... 1000,
                              stride: 5,
                              prefix: "movement",
                              getter: AnthroCharacterBody2D.getAcceleration,
                              setter: AnthroCharacterBody2D.setAcceleration)
        classInfo.registerInt(named: "speed",
                              range: 1 ... 1000,
                              prefix: "movement",
                              getter: AnthroCharacterBody2D.getSpeed,
                              setter: AnthroCharacterBody2D.setSpeed)
        classInfo.registerInt(named: "friction",
                              range: 1 ... 1000,
                              prefix: "movement",
                              getter: AnthroCharacterBody2D.getFriction,
                              setter: AnthroCharacterBody2D.setFriction)
    }()
}
```

alicerunsonfedora:
Whether these can be used for macro expansion could depend on a few factors, and it may not be necessary for macros at all. However, I do think this would be a suitable solution for now since, to my knowledge, Swift 5.9 is still in a beta phase, along with everything else introduced in WWDC. Likewise, it might be a safe bet to keep this for older Swift versions, or if people prefer to do it manually.

migueldeicaza:
I'll digest some more, they might be the building blocks the macros use, or as an option for manual configuration.

I am not too concerned about 5.9 being a requirement, as SwiftGodot itself is still in its infancy.

Sources/SwiftGodot/Extensions/ClassInfo+ConvenienceProperties.swift
```diff
@@ -0,0 +1,263 @@
...
+@available(macOS 11.0, *)
+extension UTType {
+    /// The file type that corresponds to a Godot resource file.
+    public static var godotResource = UTType(filenameExtension: "res")
+
+    /// The file type that corresponds to a Godot scene file.
+    public static var godotScene = UTType(filenameExtension: "tscn")
+
+    /// The file type that corresponds to a GDScript file.
+    public static var gdscript = UTType(filenameExtension: "gd")
```
nvanfleet, line 24: Capital "s" in Script?

alicerunsonfedora: I considered this, but it didn't feel right to me as GDScript is one word rather than "GD Script".

migueldeicaza:
Here is a taste of the macro integration I am working on, currently this works by applying the "Godot" attribute to the class, and then any method with the @Export attribute are registered in this way.

What I think I need to do is make it so that the @Export attribute can take parameters for properties that define the sort of information that is passed to the registration, so that it is not hardcoded:

https://github.com/migueldeicaza/SwiftGodot/issues/86

alicerunsonfedora:
> Here is a taste of the macro integration I am working on, currently this works by applying the "Godot" attribute to the class, and then any method with the @export attribute are registered in this way.
> 
> What I think I need to do is make it so that the @export attribute can take parameters for properties that define the sort of information that is passed to the registration, so that it is not hardcoded:
> 
> #86

It does look like an interesting approach, though I wonder if it makes more sense to split the export up to different types to make it clearer as to what's happening:

```swift
class SomeNode: Node {
    @PickerNameProvider
    enum PlayerClass: Int {
        case barbarian
        case mage
        case wizard
    }
    @ExportedEnum public var playerClass: PlayerClass = .barbarian 
}
```

Could end up producing something like:

```swift
static var classInfoRegisty: Void = {
    let classInfo = ClassInfo<SomeNode>(name: StringName("SomeNode"))
    classInfo.registerEnum(named: "playerClass",
                           for: PlayerClass.self,
                           getter: SomeNode._getPlayerClass,
                           setter: SomeNode._setPlayerClass)
}

func _getPlayerClass(args _: [Variant]) -> Variant? { Variant(playerClass.rawValue) }
func _setPlayerClass(args: [Variant]) -> Variant? {
    ClassInfo.withCheckedProperty(named: "playerClass", args: args) { arg in
        playerClass = PlayerClass(rawValue: Int(arg) ?? 0) ?? .barbarian
    }
}
```

migueldeicaza:
My general feeling is to not declare too many macros for individual exports, but this might be an exception, because one bit I have struggled with is this: macros operate at the syntax-level, they do not really operate at the semantic level.

So when I am looking at this piece of code: `public var playerClass: PlayerClass = .barbarian ` there is no easy way of knowing that `PlayerClass` is an enumeration or what their values are: as far as SwiftSyntax is concerned this is a type identifier.   So the @ExportedEnum usage would be a good hint to drive the registration in a different way.

Another option is to add arguments to @Export, so we could do various things like:

```
@Export (hint: "Measured in meters", minValue: 0, maxValue: 10)
var distance: Float
```

So in this case, it could be:

```
@Export (useEnum: true)
var public var playerClass: PlayerClass = .barbarian 
```

migueldeicaza:
Btw, I suggest we move this discussion to discussions.

alicerunsonfedora:
> Btw, I suggest we move this discussion to discussions.

Sure, I've moved the disucssion to #94 .

PR #121, opened 2023-09-24
Foundation import cleanup

tishin:
All Foundation imports in SwiftGodot target and generated sources can be removed since none of those use any Foundation functionality.
Foundation imports also complicate builds for platforms other than macOS since it's a standalone package there (and Foundation is not fully ported to Swift as of yet).


migueldeicaza:
Thank you for this patch!

PR #122, opened 2023-09-24
Clean up unnecessary dependency

tishin:
CodeGeneratorPlugin's dependency should be enough, There's no need for the core target to depend on Generator.


migueldeicaza:
Lovely, thank you for this patch!

PR #123, opened 2023-09-24
Use GDExtensionVariantType.RawValue abstraction

tishin:
GDExtensionVariantType.RawValue is Int32 when building on Windows. Using RawValue abstraction instead of concrete type should help to avoid type mismatch.


migueldeicaza:
Thank you for this patch!

PR #125, opened 2023-09-25
Wrapped: Cast generic types to AnyObject explicitly

tishin:
Trying to build SwiftGodot on Windows results in the following errors:
```
C:\Projects\SwiftGodot\Sources\SwiftGodot\Core\Wrapped.swift:197:46: error: argument type 'T.Type' expected to be an instance of a class or class-constrained type
    guard let superStr = getSuperType (type: type) else {
                                             ^
C:\Projects\SwiftGodot\Sources\SwiftGodot\Core\Wrapped.swift:170:54: error: argument type 'T.Type' expected to be an instance of a class or class-constrained type
    let retained = Unmanaged<AnyObject>.passRetained(type)
                                                     ^
```
Explicit cast to AnyObject seems to fix the issue.

### DIFF
--- a/Sources/SwiftGodot/Core/Wrapped.swift
+++ b/Sources/SwiftGodot/Core/Wrapped.swift
@@ -167,7 +167,7 @@ func register<T:Wrapped> (type name: StringName, parent: StringName, type: T.Typ
     info.get_virtual_func = getVirtual
     info.notification_func = notificationFunc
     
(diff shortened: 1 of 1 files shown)


tishin:
Building on Ubuntu also has this issue. I'm using Swift 5.9 toolchain in both cases.

migueldeicaza:
Thank you for your patch, and apologies for the delay, I missed it!

PR #126, opened 2023-09-25
Add availability checks for UniformTypeIdentifiers

tishin:
Since the framework is not available for platforms other than macOS.


migueldeicaza:
Thanks for the patch!

PR #127, opened 2023-09-26
Fix method flags type mismatch on Windows

tishin:
Don't know why, but C enums on Windows implicitly get signed integer as the underlying type, which results in type mismatch for method_flags assignment (which is explicitly declared as uint32_t). This was also the cause for type mismatch at #123. Doesn't happen when building on Ubuntu.


migueldeicaza:
Thank you so much for tracking these down!

PR #129, opened 2023-09-26
Add additional macros

alicerunsonfedora:
I'm introducing the macro library from Indexing Your Heart's source upstream since I think they will benefit developers using this package to create scripts for Godot:

- `@NativeHandleDiscarding` is a macro that initializes `init(nativeHandle:)` for you. This isn't meant to supersede the existing `@Godot` macro, but it can be used if developers prefer to do things manually.
- `#initSwiftExtension` is a macro for quickly setting up an extension by writing the entrypoint methods for you, only requiring the C name and the array of types to be registered with the Godot engine.
- `#texture2DLiteral` is a macro for quickly fetching an image resource and using it as a texture.
- `@PickerNameProvider` is a macro for giving an enum the backing and conformances needed to register them in Godot.
- `@SceneTree` is a macro that adds a `get` accessor to properties, referencing the scene tree. Consider this a replacement of the existing `@BindNode` property wrapper.

I've also updated the tutorials to utilize the new macros.

> **Note**  
> Indexing Your Heart is licensed under the CNPLv7, and these macros were a part of that in a separate package. However, because I am the sole author of the game at the moment with no contributors, I have the authority to have them relicensed under SwiftGodot's existing license. I also highly doubt they can be used to harm someone directly.


Sources/SwiftGodotMacros/MacroDefs.swift
```diff
@@ -8,7 +8,92 @@
...
+                                                                              type: "InitSwiftExtensionMacro")
+
+/// A macro that instantiates a `Texture2D` from a specified resource path. If the texture cannot be created, a
+/// `preconditionFailure` will be thrown.
+///
+/// Use this to quickly instantiate a `Texture2D`:
+/// ```swift
+/// func makeSprite() -> Sprite2D {
+///     let sprite = Sprite2D()
+///     sprite.texture = #texture2DLiteratl("res://assets/playersprite.png")
```
PadraigK, line 38:
```suggestion
///     sprite.texture = #texture2DLiteral("res://assets/playersprite.png")
```

Tests/SwiftGodotMacrosTests/TextureLiteralMacroTests.swift
```diff
@@ -0,0 +1,52 @@
...
+            """
+            let spriteTexture = #texture2DLiteral("res://assets/icon.png")
+            """,
+            expandedSource: """
+            let spriteTexture = {
+                guard let texture: Texture2D = GD.load(path: "res://assets/icon.png") else {
+                    preconditionFailure(
+                        "Texture could not be loaded.",
+                        file: "TestModule/test.swift",
+                        line: 1)
```
PadraigK, line 29: Just curious, is there a way to see this "Texture could not be loaded." text when the game hits it? Any fatalError / preconditionFailures I've tried just leads to a crash in godot, and even the crash doesn't seem to show the actual failure text.

alicerunsonfedora: Not to my knowledge, but I could call a `GD.pushError` beforehand so that the error is present in the Godot console before the crash.

PadraigK: Thanks for getting back to me! I just tried this and the error doesn't make it to the Godot console unfortunately. No worries either way, this is not a reason to change the PR, which is really cool overall! Thanks for sharing it.

alicerunsonfedora: Ah, I guess it must happen so quickly that Godot doesn't even have the chance to report the error. Though I wonder if it appears in the stacktrace or logs.

alicerunsonfedora: Appears this might require a flag of some kind:

> `message`. 
> A string to print in a playground or -Onone build. The default is an empty string.

https://developer.apple.com/documentation/swift/preconditionfailure(_:file:line:)

PadraigK: A debug build should be `-Onone` already, I suspect it's dying before the logs are flushed. I'll open a separate issue to discuss this.

PadraigK: Opened https://github.com/migueldeicaza/SwiftGodot/issues/132 to discuss.

alicerunsonfedora: @PadraigK Perhaps you can try replacing my `GD.pushError` with Swift's default `print` and see if you get something in the terminal?

PadraigK:
To get this working in my project, I had to add `.product(name: "SwiftGodotMacros", package: "SwiftGodot")` to my `.target` dependencies. It would be good to add this to the tutorial where it describes the setup of the driver package.

Sources/SwiftGodotMacroLibrary/SceneTreeMacro.swift
```diff
@@ -0,0 +1,97 @@
...
+//
+
+import Foundation
+import SwiftCompilerPlugin
+import SwiftDiagnostics
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SwiftSyntaxMacros
+
+public struct SceneTreeMacro: AccessorMacro {
```
migueldeicaza, line 15: I am still a bit concerned that we have duplicate functionality now, and would rather only have one way of doing things.

I am ok taking this over my existing solution, but we should at least make this default the path to being the name of the variable, so users do not have to repeat themselves, something like:

```swift
@SceneTree var myTimer: Timer
```

Should infer that the name of the timer in the scene is `myTimer`

migueldeicaza (approved):

PR #135, opened 2023-09-27
Remove classInfo variable

nvanfleet:
Since I build without the option of suppressing warnings I get this error when I build. This is because we don't use the classInfo variable.

```
/var/folders/vg/yjyvdt016d5_ws4fz5_qycmc0000gn/T/swift-generated-sources/@__swiftmacro_19PlantGenerationDemo26CatmullRomSplineVisualizer5GodotfMm_.swift:12:9: warning: initialization of immutable value 'classInfo' was never used; consider replacing with assignment to '_' or removing it
    let classInfo = ClassInfo<CatmullRomSplineVisualizer> (name: className)
        ^
/Users/nvanfleet/src/PlantGeneration/swift_src/Demo/CatmullRomSplineVisualizer.swift:8:7: note: in expansion of macro 'Godot' here
final class CatmullRomSplineVisualizer: Node3D {
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/Sources/SwiftGodotMacroLibrary/MacroGodot.swift
+++ b/Sources/SwiftGodotMacroLibrary/MacroGodot.swift
@@ -165,7 +165,7 @@ class GodotMacroProcessor {
     """
     static func _initClass () {
         let className = StringName("\(className)")
-        let classInfo = ClassInfo<\(className)> (name: className)\n
+        let _ = ClassInfo<\(className)> (name: className)\n
     """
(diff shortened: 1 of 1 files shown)


migueldeicaza:
The issue with the patch is that it removes the name for every scenario, so when you use @Callable or @Export it will fail.

nvanfleet:
Oh interesting. I didn't realize that.

PR #137, opened 2023-09-28
Permit implictly unwrapped optionals when using @SceneTree

PadraigK:
Since I know that I'm only going to reference these properties after they are available in `_ready()`, I prefer to implicitly unwrap the optional. At the moment a diagnostic warning prevents me from doing this. This PR loosens the requirement a bit so that we can use `!` as well as `?`. It will still emit a diagnostic if the type is missing both.

```swift
class MainScene: Node {
    @SceneTree(path: "StartPosition") var startPosition: Marker2D!
    @SceneTree(path: "Camera2D") var camera: Camera2D!
}
```


migueldeicaza:

PR #138, opened 2023-09-28
Fix process argument limit error when building on Windows

tishin:
Building SwiftGodot with full list of generated sources on Windows exceeds the 32K symbol limit of `CreateProcess` resulting in

`error: command Compiling Swift Module 'SwiftGodot' (899 sources) failed: unable to spawn process 'C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swiftc.exe' (The filename or extension is too long.)`

Merging all generated sources into a single file should fix the issue for now until it's resolved from the SPM side (or maybe modularization of sources in the project).


migueldeicaza:
Makes sense.

One possible optimization for Windows would be to split the sharedPrinter from the printer, but pass a flag indicating that the generated output is being blended.   And then concatenate the various outputs of printers at the end.

The sole reason for that is that it significantly reduces the build time.

Happy to merge in this way if you prefer.

migueldeicaza:
Thank you for the update!

PR #139, opened 2023-09-29
Fixes issue where Callable macro expansion would not register method correctly

PadraigK:
The macro was expanding to register a func with the literal name "funcName" instead of the contents of the variable `funcName`.

Also adds tests for the Godot, Export and Callable macros.

### DIFF
--- a/Tests/SwiftGodotMacrosTests/MacroGodotTests.swift
+++ b/Tests/SwiftGodotMacrosTests/MacroGodotTests.swift
@@ -1,0 +1,175 @@
+//
+//  MacroGodotTests.swift
+//  
+//
+//  Created by Padraig O Cinneide on 2023-09-28.
+//
+
+import SwiftSyntaxMacros
+import SwiftSyntaxMacrosTestSupport
+import XCTest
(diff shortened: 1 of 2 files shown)


Tests/SwiftGodotMacrosTests/MacroGodotTests.swift
```diff
@@ -0,0 +1,175 @@
...
+				@Export var goodName: String = "Supertop"
+			}
+			""",
+			expandedSource:
+            """
+            class Hi: Node {
+            	var goodName: String = "Supertop"
+            
+            	func _mproxy_set_goodName (args: [Variant]) -> Variant? {
+            		/*    goodName = String(args [0])!*/
```
PadraigK, line 139: Not changed in this PR, but surprising that this expands to a `/* */` block — is this a bug @migueldeicaza?

migueldeicaza: Yes, this was some leftover debugging code, I will remove it.

migueldeicaza:

PR #141, opened 2023-09-30
Fix source generation on Windows

tishin:
1. `buildCommand` does not work on Windows seemingly because of an SPM bug. I described the whole thing in https://github.com/apple/swift-package-manager/issues/6962. Using `prebuildCommand` instead fixes the issue.
2. After fixing codegen command Windows build fails with `You don't have permission` error. I suggest it's because generator executable doesn't have write permissions to the temporary directory `Foundation` implementation uses for atomic writes, since toggling off `atomically` fixed the issue for me. Not sure what benefit atomic writes can bring in context of code generator, so I disabled it for all platforms.

### DIFF
--- a/Generator/Generator/Printer.swift
+++ b/Generator/Generator/Printer.swift
@@ -59,7 +59,7 @@ class Printer {
                 return
             }
         }
-        try! result.write(toFile: file, atomically: true, encoding: .utf8)
+        try! result.write(toFile: file, atomically: false, encoding: .utf8)
     }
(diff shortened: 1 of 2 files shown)


tishin:
This should finally make the package buildable on Windows as is.

migueldeicaza:
I used atomically blindly, did not care either way, and I just had to pick a value :-)

Good to know that this causes problems on Windows, I will avoid in the future

PR #142, opened 2023-09-30
Make SwiftGodotMacros a dynamic library

tishin:
To match the existing linking behavior of the base library.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,6 +18,7 @@ let package = Package(
             targets: ["SwiftGodot"]),
         .library(
             name: "SwiftGodotMacros",
+            type: .dynamic,
             targets: ["SwiftGodotMacros"]),
         .plugin(name: "CodeGeneratorPlugin", targets: ["CodeGeneratorPlugin"]),
         .library(


migueldeicaza:
Thank you - this makes sense.

migueldeicaza:
Mhm, this actually became through some side effect a dependency of SwiftGodot which is needed to run things.

I will likely have to undo this later today.

tishin:
There's no side effect dependency. If you you use macros, you also need `SwiftGodotMacros` to run things. Which embeds `SwiftGodot` in itself, so the first of driver's dependencies in tutorials is redundant: 
```
.target(
    name: "SimpleRunnerDriver",
    dependencies: [
        "SwiftGodot",
        .product(name: "SwiftGodotMacros", package: "SwiftGodot")
    ],
```
It will work with a single dependency on 'SwiftGodotMacros':
```
.target(
    name: "SimpleRunnerDriver",
    dependencies: [
        .product(name: "SwiftGodotMacros", package: "SwiftGodot"),
    ],
```
Because `SwiftGodotMacros` statically depends on `SwiftGodot` target and embeds it.

Issue #144, opened 2023-10-01
Idea to improve prefix removal on enum cases


PadraigK:
In enums like PropertyUsageFlags, the current prefix removal strategy of removing the enum name from each case doesn't work, because the cases are all prefixed with "PROPERTY_USAGE_" rather than "PROPERTY_USAGE_FLAGS". 

We could improve this by identifying the common prefix of all cases and removing that, instead of relying on the enum name matching the case prefix. 

Alternatively a simpler fix might be to special case enum names containing FLAGS, since they all seem to follow the pattern of dropping the "_FLAGS" part.

PadraigK:
> Alternatively a simpler fix might be to special case enum names containing FLAGS, since they all seem to follow the pattern of dropping the "_FLAGS" part.

Actually this is not the case for `PipelineDynamicStateFlags` which has `dynamicStateLineWidth` and a number of other enums, so ignore this alternative.

PadraigK:
Here's what I've got so far: https://gist.github.com/PadraigK/af78939c3e597635d6db5dc98e7b6c64

migueldeicaza:
Closing by your PR

Issue #23, opened 2023-04-14
Bug: Enum generation produces MethodFlags.tatic instaead of MethodFlags.static


migueldeicaza:

PadraigK:
I got curious about this one and took a look. It turns out that the prefix removal is causing this, because the prefix here is "MethodFlags" and the case is "methodFlagStatic". 

This enum (MethodFlags) is a big inconsistently named on the Godot side. At the moment the default case is prefixed with METHOD_FLAGS (plural) while the rest aren't. I would argue that either the name should be singular, and default case changed to match, or the name should stay plural and the other cases should match that. 

```
METHOD_FLAG_NORMAL
METHOD_FLAG_EDITOR
METHOD_FLAG_CONST
METHOD_FLAG_VIRTUAL
METHOD_FLAG_VARARG
METHOD_FLAG_STATIC
METHOD_FLAG_OBJECT_CORE
METHOD_FLAGS_DEFAULT   
```

To resolve our bug here, we could:

1. Just special case this one string.
2. Apply the prefix removal *before* snake casing, so we don't pull in the first letter of "static" 
3. Try to land an upstream change in Godot itself to update these to be consistent. 

I'm happy to do the PR in any case, but thought I'd see if there are any preferences on the approach (or alternatives that I haven't considered)

PadraigK:
~3. Try to land an upstream change in Godot itself to update these to be consistent.~

I asked here https://github.com/godotengine/godot/issues/82612 but there's no appetite since it would be a compatibility breaking change.

PR #145, opened 2023-10-01
Change enum prefix detection strategy

PadraigK:
Fixes issue: https://github.com/migueldeicaza/SwiftGodot/issues/144

The generator will now find the common prefix among all case names instead of relying on the enum definition name.

You can see the changes to the generated API in this gist: https://gist.github.com/PadraigK/af78939c3e597635d6db5dc98e7b6c64

### DIFF
--- a/Generator/Generator/StringOperations.swift
+++ b/Generator/Generator/StringOperations.swift
@@ -23,7 +23,7 @@ extension String {
         return self.processCamalCaseRegex(pattern: acronymPattern)?
             .processCamalCaseRegex(pattern: normalPattern)?.lowercased() ?? self.lowercased()
     }
-              
+    
     fileprivate func processCamalCaseRegex(pattern: String) -> String? {
         let regex = try? NSRegularExpression(pattern: pattern, options: [])
         let range = NSRange(location: 0, length: count)
@@ -57,7 +57,7 @@ extension String {
 func escapeSwift (_ id: String) -> String {
     switch id {
     case "protocol", "func", "static", "inout", "in", "self", "case", "repeat", "default",
-         "import", "init", "continue", "class", "operator", "where", "var", "enum", "nil", "extension", "internal":
+        "import", "init", "continue", "class", "operator", "where", "var", "enum", "nil", "extension", "internal":
         return "`\(id)`"
     default:
         return id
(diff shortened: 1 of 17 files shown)


PadraigK:
This needs a few more tweaks to compile, I'll get to them later on today.

PadraigK:
Ok, this should be ready for a review now. I think this is a nice improvement in the naming of enums but it impacts a lot of stuff. 

It introduces one slight naming regression for MethodFlags cases because the inconsistent prefix on that enum means we don't identify a great prefix (more context here:  https://github.com/migueldeicaza/SwiftGodot/issues/23) — 

Relevant part of the diff:
```diff
-    public static let methodFlagNormal = MethodFlags (rawValue: 1)
-    public static let methodFlagEditor = MethodFlags (rawValue: 2)
-    public static let methodFlagConst = MethodFlags (rawValue: 4)
-    public static let methodFlagVirtual = MethodFlags (rawValue: 8)
-    public static let methodFlagVararg = MethodFlags (rawValue: 16)
-    public static let tatic = MethodFlags (rawValue: 32)
-    public static let methodFlagObjectCore = MethodFlags (rawValue: 64)
-    public static let `default` = MethodFlags (rawValue: 1)
+    public static let flagNormal = MethodFlags (rawValue: 1)
+    public static let flagEditor = MethodFlags (rawValue: 2)
+    public static let flagConst = MethodFlags (rawValue: 4)
+    public static let flagVirtual = MethodFlags (rawValue: 8)
+    public static let flagVararg = MethodFlags (rawValue: 16)
+    public static let flagStatic = MethodFlags (rawValue: 32)
+    public static let flagObjectCore = MethodFlags (rawValue: 64)
+    public static let flagsDefault = MethodFlags (rawValue: 1)
```

The impact is that the `default` case is called `flagsDefault`. I think we should just special-case this enum. I'd like to handle that separately though, but if its a blocker I can do it on this PR.

migueldeicaza:
I think it is ok to special-case the default case to get that name in place.

I also think that we should special case the keys, because I am not a fan of "_1" as an enum value vs say "key1".

PadraigK:
Sounds good, I'll get those changes in soon.

PadraigK:
Ok should be good now; anytime there would be an invalid swift name, it does the same thing it would have done before, so there is no change on those cases; e.g. we get `key1` instead of `_1`
  
Here's the diff in API: https://gist.github.com/PadraigK/af78939c3e597635d6db5dc98e7b6c64

Generator/Generator/StringOperations.swift
```diff
@@ -57,9 +57,68 @@ extension String {
...
+        }
+        let prefix = first.commonPrefix(with: last)
+        return prefix != "" ? prefix : nil
+    }
+}
+
+extension [JGodotValueElement] {
+    func commonPrefix() -> String {
+        map(\.name).commonPrefix()?.dropAfterLastUnderscore() ?? ""
+    }
```
alicerunsonfedora, line 84: Is `commonPrefix` often invoked to warrant providing this in an extension?

Generator/Generator/StringOperations.swift
```diff
@@ -57,9 +57,68 @@ extension String {
...
+}
+
+extension String {
+    func dropPrefix(_ prefix: String) -> String {
+        guard hasPrefix(prefix) else { return self }
+        guard prefix != self else { return self }
```
alicerunsonfedora, line 90:
```suggestion
        guard hasPrefix(prefix), prefix != self else { return self }
```

Generator/Generator/StringOperations.swift
```diff
@@ -57,9 +57,68 @@ extension String {
...
+        let suffix = String(dropFirst(prefixSize))
+        return suffix.isValidSwiftName() ? suffix : self
+    }
+    
+    func dropAfterLastUnderscore() -> String? {
+        if let range = range(of: "_", options: .backwards) {
+            return String(prefix(upTo: range.upperBound))
+        } else {
+            return nil
+        }
```
alicerunsonfedora, line 111:
```suggestion
        return nil
````

Generator/Generator/StringOperations.swift
PadraigK: It is used in 2 or 3 places. I did it like this to ensure consistent use at call sites, since things like default method parameter mappings need to follow the same pattern as the case definition.

alicerunsonfedora: No idea how to mark this conversation as resolved since the button to do so isn’t on my end, but consider this conversation “resolved” for all intents and purposes.

Issue #148, opened 2023-10-01
Need to copy `libSwiftGodotMacros.dylib` into `/bin` folder too


PadraigK:
Godot complains if I don't also copy this dylib into the /bin folder.

I suspect this is the cause: https://github.com/migueldeicaza/SwiftGodot/commit/f9f00d7136e4968c4315e41c174f138f8842135b

If we want to keep this dynamic, we should update the docs to instruct users to also copy this dylib.

migueldeicaza:
Well, this is odd - because this should not be necessary at runtime at all.

What is the error that you get?

tishin:
It is necessary to copy the `libSwiftGodotMacros.dylib` alongside extension's dylib, just like it was necessary to copy `libSwiftGodot.dylib` before macros were introduced. Tutorials cover that: https://github.com/migueldeicaza/SwiftGodot/blob/main/Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your%20First%20Extension.tutorial#L186
I think they should either both be static or both be dynamic. The way it worked before https://github.com/migueldeicaza/SwiftGodot/commit/f9f00d7136e4968c4315e41c174f138f8842135b: 
- when using plain `SwiftGodot`, its `libSwiftGodot.dylib` was linked dynamically (and had to be copied separately from driver's dylib)
- when using `SwiftGodotMacros`, the library (both `SwiftGodot` and `SwiftGodotMacros` targets) got embedded into driver's dylib (and the driver's dylib was enough to run the extension, contradicting tutorials)

Which I thought was confusing.

tishin:
The way it works now:
- when using `SwiftGodot`, `libSwiftGodot.dylib` has to be placed alongside driver's dylib
- when using `SwiftGodotMacros`, `libSwiftGodotMacros.dylib` has to be placed alongside driver's dylib

tishin:
Oh, I see the point of confusion now. You're probably using an extra dependency on `SwiftGodot` like so:
```
.target(
    name: "SimpleRunnerDriver",
    dependencies: [
        "SwiftGodot",
        .product(name: "SwiftGodotMacros", package: "SwiftGodot")
    ],
```
[...]

PR #150, opened 2023-10-03
Move macro definitions into SwiftGodot target

tishin:
This should avoid all the complications discussed in #148, leaving `SwiftGodot` the only dependency and only import to worry about for the users. And it'll keep the initial composition of `libDriver.dylib`+`libSwiftGodot.dylib` as the build output.

But it'll require additional workarounds for Windows, since macros aren't supported there yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,4 +22,4 @@ check-args:
 	@if test x$(VERSION)$(NOTES) = x; then echo You need to provide both VERSION=XX NOTES=FILENAME arguments to this makefile target; exit 1; fi
 
 sync:
-	@if test ../SwiftGodotBinary; then rsync -a Sources/SwiftGodotMacros Sources/SwiftGodotMacroLibrary ../SwiftGodotBinary/Sources; else echo "missing directory ../SwiftGodotBinary"; fi
+	@if test ../SwiftGodotBinary; then rsync -a Sources/SwiftGodotMacroLibrary ../SwiftGodotBinary/Sources; else echo "missing directory ../SwiftGodotBinary"; fi
--- a/Package.swift
+++ b/Package.swift
@@ -24,10 +24,6 @@ let package = Package(
             name: "SwiftGodot",
             type: .dynamic,
             targets: ["SwiftGodot"]),
-        .library(
-            name: "SwiftGodotMacros",
-            type: .dynamic,
-            targets: ["SwiftGodotMacros"]),
         .plugin(name: "CodeGeneratorPlugin", targets: ["CodeGeneratorPlugin"]),
         .library(
             name: "SimpleExtension",
@@ -70,7 +66,7 @@ let package = Package(
             dependencies: ["GDExtension"],
             swiftSettings: [.unsafeFlags (["-suppress-warnings"])],
             linkerSettings: linkerSettings,
-            plugins: ["CodeGeneratorPlugin"]),
+            plugins: ["CodeGeneratorPlugin", "SwiftGodotMacroLibrary"]),
         
         // These are macros that can be used by third parties to simplify their
         // SwiftGodot development experience, these are used at compile time by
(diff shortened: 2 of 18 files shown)


alicerunsonfedora:
I don't think we should merge the two targets together because then that would imply that you would need both to operate. Additionally, this can become much harder to test, and I imagine this could break how macros work in the future should Swift change how it operates. Rather, we should look into creating a new target that imports both so that you can still have the single import statement while also having the smaller, more composable targets.

tishin:
But currently you do need macro definitions target `SwiftGodotMacros` (not macro implementations target `SwiftGodotMacroLibrary`) to operate. It was just hidden by static dependency in the initial implementation.

Look at the case with static dependency on `SwiftGodotMacros` in #148.
Why would driver's dylib jump in size from 100 Kb to 70 Mb after adding macros? Because driver's dylib also embedded `SwiftGodotMacros` target due to static dependency, which itself depended on `SwiftGodot` statically, making driver also embed `SwiftGodot`.
Why don't you need to copy `libSwiftGodot.dylib` to the `/bin` directory anymore after implementing macros? Same thing, because both `SwiftGodot` and `SwiftGodotMacros` are already in the driver's dylyb due to static dependency.

I'm fine with the static solution and embedding everything into a single dylib if that's what everyone prefers, just make `SwiftGodot` product also static to match it.

alicerunsonfedora:
> But currently you do need macro declarations target `SwiftGodotMacros` (not macro implementations target `SwiftGodotMacroLibrary`) to operate. It was just hidden by static dependency in the initial implementation.
> 
> Look at the case with static dependency on `SwiftGodotMacros` in #148. Why would driver's dylib jump in size from 100 Kb to 70 Mb after adding macros? Because driver's dylib also embedded `SwiftGodotMacros` target due to static dependency, which itself depended on `SwiftGodot` statically, making driver also embed `SwiftGodot`. Why don't you need to copy `libSwiftGodot.dylib` to the `/bin` directory anymore after implementing macros? Same thing, because both `SwiftGodot` and `SwiftGodotMacros` are already in the driver's dylyb due to static dependency.
> 
> I'm fine with the static solution and embedding everything into a single dylib if that's what everyone prefers, just make `SwiftGodot` product also static to match it.

Ideally, you can/should be able to still use the SwiftGodot project _without using the macros_. My issue here is that we're now bundling everything together in this singular, monolithic package with extra bloat, and it's no longer modular. By making the macros a core part of SwiftGodot in order for it to function, we lose this capability.

In my testing and usage, I've always had to copy over libSwiftGodot.dylib (or, in my case, SwiftGodot.framework), and my binary sizes have never ballooned up to 70+ MB. I suspect the reasoning as to why the file sizes are getting larger is _because_ everything is being bundled together into a single package, rather than being broken up into modular, composable parts.

image

> 1. Single "import SwiftGodot" for users, that allows them to use the API and the macros (see the threads I posted on the forums.swift.org, that sadly have gone unanswered)

I've found a way to do this that works really well and still keeps the modular structure, which you can see at https://gitlab.com/indexing-your-heart/engine/swiftgodot. I fear that combining everything into a single structure will also affect the build sizes and other quirks for the modular system I have set up in place, and it will take considerable effort to keep it modularized.

migueldeicaza:
> In my testing and usage, I've always had to copy over libSwiftGodot.dylib (or, in my case, SwiftGodot.framework), and my binary sizes have never ballooned up to 70+ MB. I suspect the reasoning as to why the file sizes are getting larger is because everything is being bundled together into a single package, rather than being broken up into modular, composable parts.

I went and tried this patch, for both an extension and the general purpose library, and the size difference are as follows:

|  | Current Code | With this Patch Applied |Difference|
|--------|-------:|------:|-----:|
| SimpleExtension | 60,217,376 | 60,216,048 |1,328|
| SwiftGodot | 60,117,176 | 60,117,176 |0|

So the growth is marginal, I do not see any problems with this.

I am happy to have a separate discussion on how to reduce the executable size, and other strategies that we could use to cope with the growth over time, but last night 20-minute hack reduced the library by about six megabytes, so unless there are stronger objections, I think that this patch achieves the ease of use goals that I want to get to.

SIDEBAR: the ballooning of the library size is driven to a large extent to the API surface changes from 4.0 to 4.1, and in part by some code generation changes.

alicerunsonfedora:
> I am happy to have a separate discussion on how to reduce the executable size, and other strategies that we could use to cope with the growth over time, but last night 20-minute hack reduced the library by about six megabytes, so unless there are stronger objections, I think that this patch achieves the ease of use goals that I want to get to.

I still don’t think it’s a good idea to have everything merged into a single monolithic binary like this, and I don’t think having a single import statement is worth throwing out the current modularity. Sure, having a single import statement is easy; however, simple and easy aren’t necessarily the same thing. Loris Cro does a good job at explaining the differences here: https://kristoff.it/blog/simple-not-just-easy/

I do believe we can strike a good balance between simple and easy, but I don’t think merging everything into a single binary with extra bloat developers didn’t ask for is the most efficient way of doing so.

migueldeicaza:
The macro implementations are not part of this library, only the macro definitions are, and they are an integral part of the experience that I would like to expose to users, because the alternative is riddled with boiler plate and things that are too easy to get wrong.

I would have loved to have explored these alternatives using the reflection proposal (which would increase the executable size much more than these macros do), but that proposal seems to be stuck, while the macros are not.

migueldeicaza:
An additional bonus to the points 1..4 before:

For the first time since macros were introduced, consuming this library from Xcode projects, or standalone SwiftPM projects work.   I am beyond psyched about it.

Issue #151, opened 2023-10-03
Error: Attempt to register extension class using non-existing parent class 'NSObject'


lupuchard:
I'm doing the "Your First Extension" tutorial and trying to adapt it to Linux. When Godot tries to load the extension it gives this error:
```
SWIFT: extension_initialize
PlayerController
NSObject
ERROR: Attempt to register an extension class 'PlayerController' using non-existing parent class 'NSObject'
```

But my PlayerController is inheriting from CharacterBody2D as the tutorial has it:
```
import Foundation
import SwiftGodot
import SwiftGodotMacros

@Godot
class PlayerController: CharacterBody2D {
    var acceleration: Float = 100
    var friction: Double = 100
    ...
```

The only changes I made were for the platform in SimpleRunnerDriver.gdextension and omitting "platforms: [.macOS(.v13)]" in Package.swift.

tishin:
There's an issue with superclass reflection in `SwiftGodot`. I made a simplified example of the current implementation:

```
class Wrapped {}
class Parent: Wrapped {}
class Child: Parent {}

func getSuper<T: Wrapped>(of type: T.Type) -> String? {
    return getSuperType(type: type as AnyObject)
}

func getSuperType (type: AnyObject) -> String? {
    guard let t = Mirror (reflecting: type).superclassMirror?.subjectType else {
        return nil
    }
    return String (describing: t)
}
```

And tested it with
```
XCTAssertEqual(getSuper(of: Child.self), "Module.Parent")
XCTAssertEqual(getSuper(of: Parent.self), "Module.Wrapped")
```
The test passes on macOS, but fails on both Windows and Linux (both classes get `NSObject` as their superclass). I suppose casting generic types to AnyObject on macOS involves some ObjC interop, which can't be done on other platforms.

migueldeicaza:
This is very odd, I do not have any references to NSObject in mycodebase, I wonder what could possibly be it.

migueldeicaza:
@tishin Thank you for tracking this down!

Would you mind trying this version here?   I guess I ran into this issue in the previous incarnation of Swift Godot and had shared this answer:

https://stackoverflow.com/a/66354091/16929

migueldeicaza:
This should be fixed with the PR above

tishin:
Yes, using `Swift._getSuperclass` seems to fix the type registration issue.

But now I'm concerned about this other cast:
[...]

PR #156, opened 2023-10-06
Fix superclass reflection

tishin:
This should address #151


migueldeicaza:
Thanks for taking this for a spin!

Issue #157, opened 2023-10-06
Calling `isActionPressed` on an InputEvent instance crashes Godot


PadraigK:
```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   Godot                         	       0x108e13be4 0x104cb8000 + 68533220
1   Godot                         	       0x108e13b78 0x104cb8000 + 68533112
2   Godot                         	       0x108a2c01c 0x104cb8000 + 64438300
3   Godot                         	       0x108a4f064 0x104cb8000 + 64581732
4   libSwiftGodot.dylib           	       0x1295a299c InputEvent.isActionPressed(action:allowEcho:exactMatch:) + 1184 (InputEvent.swift:100)
5   libSimpleRunnerDriver.dylib   	       0x1131be654 PlayerController._input(event:) + 140 (PlayerController.swift:27)
6   libSwiftGodot.dylib           	       0x129951be4 _Node_proxy_input(instance:args:retPtr:) + 1092 (Node.swift:2115)
7   libSwiftGodot.dylib           	       0x129951798 @objc _Node_proxy_input(instance:args:retPtr:) + 12
8   Godot                         	       0x106ecbc9c 0x104cb8000 + 35732636
9   Godot                         	       0x106ef8438 0x104cb8000 + 35914808
```

```swift
override func _input(event: InputEvent) {
        event.isActionPressed(action: "hi")
    }
```

PadraigK:
Note that calling `Input.isActionPressed(action: "hi") works fine as a work around

tishin:
There's an issue with reference handling in generated `Node.swift`:
```
let resolved_0 = gi.ref_get_object (args [0]!.load (as: UnsafeRawPointer.self))!
swiftObject._input (event: lookupLiveObject (handleAddress: resolved_0) as? InputEvent ?? InputEvent (nativeHandle: resolved_0))
```
As far as I can see, it expects the argument to be a pointer to `Ref<InputEvent>`, but actually gets a pointer to `InputEvent` itself, so I managed to fix it by simplifying argument handling:
```
let resolved_0 = args [0]!.load (as: UnsafeRawPointer.self)
swiftObject._input (event: lookupLiveObject (handleAddress: resolved_0) as? InputEvent ?? InputEvent (nativeHandle: resolved_0))
```
For some reason all refcounted objects get this wrapper in Generator:
```
if cmap.isRefcounted {
[...]

PR #161, opened 2023-10-09
Fix argument reference handling

tishin:
Should address #157, but this PR has to be taken with caution. I don't know what was the original case for handling refcounted classes separately.


migueldeicaza:
I discussed the rationale for merging this fix in issue #157, I believe now that this is correct.

Thank you for debugging this and sending a PR.

Issue #154, opened 2023-10-05
Regression: Updating array values is broken


nvanfleet:
I haven't gone back to this project in a while, but I noticed that https://github.com/migueldeicaza/SwiftGodot/issues/90 is broken again in the seemingly the same way. 

![Screenshot 2023-10-05 at 4 17 47 PM](https://github.com/migueldeicaza/SwiftGodot/assets/1486399/4c8bebae-32d9-47e7-a038-d294b5e85c6e)

This should be the same as the first image in the issue, a flat plane mesh. It's still being setup in the same was as always. It's just that I built it using the latest SwiftGodot

migueldeicaza:
Ouch.

Sadly, I do not think the previous fix is the same fix for this, I will investigate.

tishin:
This should be fixed in `main` now.

PR #163, opened 2023-10-09
Fix vector3 index operator

tishin:
This should address #154


migueldeicaza:
Thank you so much for looking into this.

PR #164, opened 2023-10-09
Update inits in macro tests

tishin:
This should fix CI checks


migueldeicaza:
Thank you once again for your help!

PR #176, opened 2023-10-15
Limit struct stored property generation to members with offsets

### LINKED ISSUE #173: `Rect2i.end` always returns an empty rect, works in GDScript
```
# GDScript
var rect = Rect2i(0, 0, 100, 100)
print(rect.end) 
# → (100, 100)
```

```
// Swift
let rect = Rect2i(x: 0, y: 0, width: 100, height: 100)
GD.print(rect.end)
// → Vector2i(x: 0, y: 0)
```

tishin:
Should fix #173. This PR should also make `Color` and `Plane` generation work fine out of the box (unless you want `red` instead of the documented `r`, in that case I'll revert it).

### DIFF
--- a/Generator/Generator/BuiltinGen.swift
+++ b/Generator/Generator/BuiltinGen.swift
@@ -138,18 +138,8 @@ func generateBuiltinCtors (_ p: Printer,
             
             // We need to initialize some variables before we call
             if let members {
-                if bc.name == "Color" {
-                    p ("self.red = 0")
-                    p ("self.green = 0")
-                    p ("self.blue = 0")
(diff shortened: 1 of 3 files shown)


tishin:
Also some of the generated inits should be more appropriate now.

migueldeicaza:
Thank you for taking care of this change.

That said, I would love to keep the names red, green and blue, perhaps as a special case.

The reason for that is purely to match the other usage idioms in other Swift frameworks.

Let me see if I can special case it.

migueldeicaza:
I did a little tune up after that.

Issue #175, opened 2023-10-15
Can't reference SwiftGodot using `.package(url: … from: 0.21.0` 


PadraigK:
If I add SwiftGodot as a dependency pointing to `branch: "main"` everything works, but if I use the `.package(url: "https://github.com/migueldeicaza/SwiftGodot", from: "0.21.0")` syntax, then `swift package update` fails with:

```
> error: the target 'SwiftGodot' in product 'SwiftGodot' contains unsafe build flags
```

I reviewed the commits between the `0.21.0` tag and `main` and they do not change anything in the Package file. I looked into it and it turns out that [Swift PM handles these two differently](https://forums.swift.org/t/question-about-contains-unsafe-build-flags-in-spm/47677).

This means that we will either need to get rid of the unsafe flags in the Swift Godot package, or we will not be able to use release tags.

--- 

The unsafe flags, for reference:
```
.unsafeFlags([
    "-Xlinker", "-undefined",
    "-Xlinker", "dynamic_lookup",
])
```

alicerunsonfedora:
I've been running into this as well, and I believe I documented it in an earlier issue (#52, I think?). A workaround I've been using is referencing the commit hash of the tag in question, rather than the tag itself

PadraigK:
Right, that works, but we ideally ought to have standard release conventions, so I think it's important to figure this out.

migueldeicaza:
This is tentatively working now with the suggested approach

PR #177, opened 2023-10-15
Remove `unsafeFlags` from Package.swift

PadraigK:
I removed this locally and did some testing — everything seems fine without it.

Fixes https://github.com/migueldeicaza/SwiftGodot/issues/175

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,14 +4,6 @@
 import CompilerPluginSupport
 import PackageDescription
 
-var linkerSettings: [LinkerSetting] = []
-#if os(macOS)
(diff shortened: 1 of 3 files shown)


PadraigK:
My formatter ran on the Package.swift so the diff is bigger than it should be. The interesting changes here are:

https://github.com/migueldeicaza/SwiftGodot/pull/177/commits/90d391009021bc061fbb3a160d6905b7346cf199

migueldeicaza:
Ok, applied the patch manually as a comment instead, and updated the docs.

PadraigK:
Thanks!

PR #182, opened 2023-10-16
Adds documentation for building on Windows

lorenalexm:
A first pass on a document outlining what is needed to build a project on the Windows platform.


migueldeicaza:
We just recently dropped the requirement for linker flags, so that part is no longer necessary

lorenalexm:
Thanks for the heads up! I have removed that section from the document.

migueldeicaza:
Thank you for updating the docs!

PR #186, opened 2023-10-18
Add memory layout tests for builtin structs

tishin:
### Memory layout tests
This PR adds a set of tests comparing memory layout of Swift structs to layout metadata from `extension_api.json`. Each test checks if:
- Swift struct size matches `builtin_class_sizes`
- Each Swift stored property's offset matches `builtin_class_member_offsets`
- Entire memory layout of the struct was covered by the test
- All members from `builtin_class_member_offsets` for the class were checked

### ExtensionApi separation
In order to reuse JSON API data models from Generator I moved the models and the JSON itself into a separate directory and separate targets:
- `extension_api.json` is not under `SwiftGodot` directory anymore, which was strange in the first place. So the source ignore of JSON is not needed.
- `ApiJsonModel.swift` is regenerated in order to make everything public. `quicktype` seems to have made a few changes since the last generation resulting in much less boilerplate code.
- `ApiJsonModel.swift` is separated from manually written additions, which are in `ApiJsonModel+Extra.swift` now. The autogenerated part is not entirely copy-pastable yet, I still had to do some tweaks around `JGodotArgument`/ `JGodotSingleton` and `JGodotGlobalEnumElement`/`JGodotBuiltinClassEnum`, but it should be easier to manage in case of future regenerations.
- `ExtensionApi` is exposed as a package product, which would allow users to write their own analyzers/validators/etc using GDExtension metadata.

### Misc
While running tests I noticed that test target `SwiftGodotMacroTests` doesn't match its directory `SwiftGodotMacrosTests`, so I fixed the target's name.


migueldeicaza:
Thank you very much!   This is a really nice contribution for the long-term maintainability of the project!

PR #188, opened 2023-10-19
Update ClassInfo methods, add InspectableProperty

alicerunsonfedora:
A new `ClassInfo.registerNodePath` method has been added to easily register NodePaths in the inspector.

`ClassInfo.registerTextView` had a typo in its signature, so this has been corrected.

The hint for a text field is now set to 'none', instead of a type, which should resolve crashes from using `ClassInfo.registerTextField`.

Finally, a new `InspectableProperty` struct has been added to house getter and setter methods more efficiently. The `@Export` macro can utilize this to create a peer definition, similar to what the `@Autovariant` and `@AutovariantEnum` macros do in SwiftGodot.

Pulled from [SwiftGodotCore](https://gitlab.com/Indexing-Your-Heart/engine/SwiftGodotCore):

- `dcf632c2dfa2dcbf784f4e78103723897ded9c80`
- `7da78a6bf1efe61be87aaa55934dbc403be59e21`
- `90a535f07f822f031d3ef214c9e3e2899b0d0afc`
- `501032ad66f844896d4b48c6b538c1987ec4b42f`
- `fc2a220da16bdc84ce31b950541d260de596016f`

**Update**

This PR also contains a fix for the macOS framework to help preserve its file structure, including symlinks, by using `cp -af` instead of `cp -rf`.


Sources/SwiftGodot/Core/InspectableProperty.swift
```diff
@@ -0,0 +1,32 @@
...
+
+import Foundation
+
+/// A structure that houses a property that can be added to a Godot inspector.
+public struct InspectableProperty<T> {
+    /// A typealias for the the method type used to register a property to Godot for inspectors.
+    public typealias RegisteredPropertyFunction = (T) -> ([Variant]) -> Variant?
+
+    /// The host object the property derives from.
+    public var hostObject: T.Type
```
PadraigK, line 16: You might not need to store this on an instance, since it'll already be in the type (e.g. it'll be `InspectableProperty<Int>.T` )

alicerunsonfedora: I believe I've tried this before with the stuff I've done with it, and it didn't quite work as intended. Hence why I have host object as a specific property in here.

Sources/SwiftGodot/Core/InspectableProperty.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    /// The getter function that gets the property.
+    public var getter: RegisteredPropertyFunction
+
+    /// The setter function that sets the property.
+    public var setter: RegisteredPropertyFunction
```
PadraigK, line 22: Could this and `getter` be `let` instead, then nothing here will be mutable.

PadraigK (commented):
LGTM!

Sources/SwiftGodot/Core/InspectableProperty.swift
alicerunsonfedora: Sure, that can be arranged.

Sources/SwiftGodot/Core/InspectableProperty.swift
PadraigK: Can you say more about what went wrong? I'm using it like this elsewhere so it'd be good to know if something can go wrong here!

Sources/SwiftGodot/Core/InspectableProperty.swift
PadraigK: `hostObject` too, if it's sticking around

Sources/SwiftGodot/Core/InspectableProperty.swift
alicerunsonfedora: I believe it was when I was trying to make the #autoProperty macro (ex.: `#autoProperty(Foo.self, "nerds"`). #autoProperty needs the variant's host type, and I couldn't fetch it from the parent class (since the parent declaration would be nil. Thus, I needed a location to store the host type into.

alicerunsonfedora: I don't imagine this will be an issue for upstream as much, since #autoProperty isn't a thing here, though.

migueldeicaza:
I like the fixes.

Couple of comments:

* I do not understand the purpose of the `InspectableProperty`, what the usage scenario is or how it is used.   This is a good opportunity to put this in the documentation for the class.   Without the sample code, or the background I am at a bit of a loss.

* With the above background, @PadraigK 's comments make me wonder what is the use of storing the type?

alicerunsonfedora:
> * I do not understand the purpose of the `InspectableProperty`, what the usage scenario is or how it is used.   This is a good opportunity to put this in the documentation for the class.   Without the sample code, or the background I am at a bit of a loss.

I plan to add overloads for the convenience methods later, but `InspectableProperty` was designed to be a simplified container for storing the getter and setter methods of a specified property. In my fork, this type is the fundamental basis for the `#autoProperty` macro, which returns the getters and setters more easily:

```swift
let someProperty = #autoProperty(MyNode.self, "rotation")
```

This is also used in SwiftGodotInspection to allow users to grab these properties more easily, without needing to supply each:

```swift
extension AnthroCharacterBody2D: GodotInspectable {
    public typealias T = AnthroCharacterBody2D
    public static var inspector: Inspector<T> {
        Inspector {
            Picker("character",
                   for: Character.self,
                   property: #autoProperty(object: AnthroCharacterBody2D.self, "character"))
            Group<AnthroCharacterBody2D>("Movement", prefix: "movement_") {
                NumberRange("speed",
                            limit: 1 ... 1000,
                            stride: 5,
                            property: #autoProperty(object: AnthroCharacterBody2D.self, "speed"))
                NumberRange("acceleration",
                            limit: 1 ... 1000,
                            property: #autoProperty(object: AnthroCharacterBody2D.self, "acceleration"))
                NumberRange("friction",
                            limit: 1 ... 1000,
                            property: #autoProperty(object: AnthroCharacterBody2D.self, "friction"))
            }
        }
    }
}
```

I imagined that this _might_ make existing macros such as @Export a little easier to work with, since one can now create a peer definition with this information pre-stored.

> With the above background, @PadraigK 's comments make me wonder what is the use of storing the type?

Storing the type was necessary for the `#autoProperty` macro to work, since I needed a clear way to get the specified type and its getters/setters (other methods such as trying to get T's type on its own didn't work really well).

---

In any case, if `InspectableProperty` is of no use to this project, I'll remove it from the PR. I had figured that, since my work uses this for things such as `SwiftGodotInspection`, that it might be beneficial to port it upstream as well.

PR #190, opened 2023-10-21
Adds #signal Macro 

PadraigK:
Adds a new freestanding macro that can be used inside a `@Godot` class to register a signal with Godot. 

Signal arguments can be defined and their types will be enforced on new `emit(signal:_argument)` methods.

```swift
@Godot 
class Player: Node2D {
    #signal("game_started")
    #signal("lives_changed", argument: ["new_lives_count": Int.self])

    func startGame() {
       emit(Player.gameStarted)
       emit(Player.livesChanged, 5)
    }
}

class Level: Area2D {
    func _ready() { 
       player.connect(Player.gameStarted, to: self, method: "game_started")
    }
    @Callable func game_started() { 
       GD.print("got game started signal!")
    }
}
```

This all works nicely with the Godot editor too — signals are exposed there and can be connected to GDScripts and well use the argument names provided in the `#signal` invocation. 

---

I did have to change `GodotVariant`'s `init()` requirement to be an `unwrap` so that subclasses of Godot objects could be made to conform to it. This makes the PR a bit more wide-reaching, but it means that it is possible to send a Node object as an argument to a `@Callable` now.

### DIFF
--- a/Generator/Generator/BuiltinGen.swift
+++ b/Generator/Generator/BuiltinGen.swift
@@ -554,9 +554,11 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
                     }
                 }
             }
+            let gtype = gtypeFromTypeName (bc.name)
+            
+            p("public static var gType: Variant.GType { .\(gtype) }")
             
             p ("/// Creates a new instance from the given variant if it contains a \(typeName)")
-            let gtype = gtypeFromTypeName (bc.name)
             // Now generate the variant constructor
             if kind == .isClass {
                 p ("public required init? (_ from: Variant)") {
--- a/Generator/Generator/BuiltinGen.swift
+++ b/Generator/Generator/BuiltinGen.swift
@@ -561,28 +561,19 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
             p ("/// Creates a new instance from the given variant if it contains a \(typeName)")
             // Now generate the variant constructor
             if kind == .isClass {
-                p ("public required init? (_ from: Variant)") {
-                    p ("guard from.gtype == .\(gtype) else") {
-                        p ("return nil")
-                    }
+                p("public static func unwrap(variant: Variant) -> \(typeName)?") {
+                    p("guard variant.gtype == .\(gtype) else { return nil }")
                     p ("var localContent: \(typeName).ContentType = \(typeName).zero")
-                    p ("from.toType(.\(gtype), dest: &localContent)")
-                    p ("// Replicate the constructor, because of a lame Swift requirement")
-                    p ("var args: [UnsafeRawPointer?] = []")
-                    p ("withUnsafePointer (to: &localContent)", arg: " ptr in") {
-                        p ("args.append (ptr)")
-                        p ("\(typeName).constructor1 (&content, &args)")
-                    }
+                    p ("variant.toType(.\(gtype), dest: &localContent)")
+                    p ("return \(typeName)(content: localContent)")
                 }
             } else {
-                p ("public init? (_ from: Variant)") {
-                    p ("guard from.gtype == .\(gtype) else") {
-                        p ("return nil")
-                    }
-                    p ("var v = \(bc.name)()")
-                    p ("from.toType(.\(gtype), dest: &v)")
-                    p ("self.init (from: v)")
-                }                
+                p("public static func unwrap(variant: Variant) -> Self?") {
+                    p("guard variant.gtype == .\(gtype) else { return nil }")
+                    p("var value = \(bc.name)()")
+                    p("variant.toType(.\(gtype), dest: &value)")
+                    p("return value")
+                }
             }
             p ("/// Wraps this \(typeName) into a Variant")
             p ("public func toVariant () -> Variant ") {
--- a/Sources/SwiftGodot/Core/ClassServices.swift
+++ b/Sources/SwiftGodot/Core/ClassServices.swift
@@ -201,7 +201,7 @@ public struct PropInfo {
     /// The name for the property
     public let propertyName: StringName
     /// The class name of the object if `propertyType` == `.object`
-    public let className: StringName?
+    public let className: StringName
     /// Property Hint for this property
     public let hint: PropertyHint
     /// Human-readable hint
@@ -209,7 +209,7 @@ public struct PropInfo {
     /// Describes how the property can be used.
     public let usage: PropertyUsageFlags
     
-    public init(propertyType: Variant.GType, propertyName: StringName, className: StringName?, hint: PropertyHint, hintStr: GString, usage: PropertyUsageFlags) {
+    public init(propertyType: Variant.GType, propertyName: StringName, className: StringName, hint: PropertyHint, hintStr: GString, usage: PropertyUsageFlags) {
         self.propertyType = propertyType
         self.propertyName = propertyName
         self.className = className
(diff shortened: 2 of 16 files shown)


migueldeicaza:
I love this change.

Here are some thoughts that I have:

* I am a bit sad that the constructor requirement is dropped, because it was so convenient to create objects from a variant.   On the other hand, I do not think we can afford to have two sets of public methods (because Windows has a limit on public methods, and I just spent a lot of time reducing our surface).    Perhaps we could just change the method name from `unwrap(variant:)` to something shorter.   Not sure.

* Now that we are blessing public Swift types with methods from the extension, I think we might need to change `gType` to be `godotType` across the board.

Here are the Api changes for this PR:

https://gist.github.com/migueldeicaza/7c0624034196453574075d375a51ec2e

Generator/Generator/BuiltinGen.swift
```diff
@@ -554,33 +554,26 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
...
-                    p ("var v = \(bc.name)()")
-                    p ("from.toType(.\(gtype), dest: &v)")
-                    p ("self.init (from: v)")
-                }                
+                p("public static func unwrap(variant: Variant) -> Self?") {
+                    p("guard variant.gtype == .\(gtype) else { return nil }")
+                    p("var value = \(bc.name)()")
+                    p("variant.toType(.\(gtype), dest: &value)")
+                    p("return value")
+                }
```
alicerunsonfedora, line 576: Perhaps this might be me, but I feel that the initializer was more appropriate in this context, and not specifically for type casting reasons. Rather, because we are constructing something with that variant. `unwrap` would imply to me that Variant is actually a representation of an optional, which, I think, might be a little confusing.

Sources/SwiftGodot/Core/ClassServices.swift
```diff
@@ -227,7 +227,8 @@ public struct PropInfo {
                         class_name: classNamePtr,
                         hint: UInt32 (hint.rawValue),
                         hint_string: hintStrPtr,
-                        usage: UInt32 (usage.rawValue))
+                        usage: UInt32 (usage.rawValue)
+                    )
```
alicerunsonfedora, line 231: Is there a particular reason why the closing parens was moved here?

Sources/SwiftGodot/Core/SignalRegistration.swift
```diff
@@ -0,0 +1,392 @@
+//
+//  SignalRegistration.swift
+//
+//
+//  Created by Padraig O Cinneide on 2023-10-18.
+//
+
+/// Describes a signal and its arguments.
+/// - note: It is recommended to use the #signal macro instead of using this directly.
+public struct SignalWithNoArguments {
```
alicerunsonfedora, line 10: I'm noticing that `SignalWithNArguments` feels ergonomically similar to that of SwiftUI's TupleView. Perhaps we can glean from that, so that we're not manually writing all these cases out?

PadraigK:
> I love this change.
> 
> 
> 
> Here are some thoughts that I have:
> 
> 
> 
> * I am a bit sad that the constructor requirement is dropped, because it was so convenient to create objects from a variant.   On the other hand, I do not think we can afford to have two sets of public methods (because Windows has a limit on public methods, and I just spent a lot of time reducing our surface).    Perhaps we could just change the method name from `unwrap(variant:)` to something shorter.   Not sure.


Yeah I think you and @alicerunsonfedora are right to call this out. I'm not super happy with unwrap either. I have a feeling that GodotVariant is now be attempting to be too many things. I will think about it some more and explore alternatives. 


> * Now that we are blessing public Swift types with methods from the extension, I think we might need to change `gType` to be `godotType` across the board.

Good point, I can definitely change this.

Thanks for the feedback!

migueldeicaza:
Apologies folks, I accidentally merged this patch, and I have now undone the push.

I think we can reopen this.

PadraigK:
I'll repost soon.

Noting some other feedback from a friend: a member macro might be more appropriate that a freestanding one, so I'll also look into that.

migueldeicaza:
One small observation, the `unwrap` method is not calling the `init(value:)` method for builtins like it did before:

```
diff -ruN SwiftGodot-REF/generated-builtin/Basis.swift SwiftGodot-DEBUG/generated-builtin/Basis.swift
--- SwiftGodot-REF/generated-builtin/Basis.swift        2023-10-21 12:11:09
+++ SwiftGodot-DEBUG/generated-builtin/Basis.swift      2023-10-21 12:12:40
@@ -13,15 +13,13 @@
 /// For more information, read the "Matrices and transforms" documentation article.
 /// 
 public struct Basis: Equatable, Hashable, GodotVariant {
+    public static var gType: Variant.GType { .basis }
     /// Creates a new instance from the given variant if it contains a Basis
-    public init? (_ from: Variant) {
-        guard from.gtype == .basis else {
-            return nil
-        }
-        
-        var v = Basis()
-        from.toType(.basis, dest: &v)
-        self.init (from: v)
+    public static func unwrap(variant: Variant) -> Self? {
+        guard variant.gtype == .basis else { return nil }
+        var value = Basis()
+        variant.toType(.basis, dest: &value)
+        return value
     }
```

migueldeicaza:
Another question, you mentioned:

> I did have to change GodotVariant's init() requirement to be an unwrap so that subclasses of Godot objects could be made to conform to it. This makes the PR a bit more wide-reaching, but it means that it is possible to send a Node object as an argument to a @Callable now.

I could not figure out how Object or Wrapper are made to conform to `GodotVariant` on this patch, did I miss something?

There are no conformances to it that I could find.

Sources/SwiftGodot/Core/ClassServices.swift
PadraigK: Oh, just a personal code formatting habit that leaked in. I'd love to start using something like swiftformat on this project some day, we've got a lot of styles mixing here.

For now I'll revert this part because it doesn't belong in the PR.

Sources/SwiftGodot/Core/SignalRegistration.swift
PadraigK: On the surface, I don't see a way that can work and still give us the nice `emit` functions that enforce the types, but if you can say more about what you have in mind I can take another look.

PadraigK:
> One small observation, the `unwrap` method is not calling the `init(value:)` method for builtins like it did before:
> 
> ```
>  public struct Basis: Equatable, Hashable, GodotVariant {
> +    public static var gType: Variant.GType { .basis }
>      /// Creates a new instance from the given variant if it contains a Basis
> -    public init? (_ from: Variant) {
> -        guard from.gtype == .basis else {
> -            return nil
> -        }
> -        
> -        var v = Basis()
> -        from.toType(.basis, dest: &v)
> -        self.init (from: v)
> +    public static func unwrap(variant: Variant) -> Self? {
> +        guard variant.gtype == .basis else { return nil }
> +        var value = Basis()
> +        variant.toType(.basis, dest: &value)
> +        return value
>      }
> ```

Hmm, I see the difference but I don't quite follow why it should call the initializer? 

1. We've made a new instance with `var value = Basis()`
2. Then copied the variant's storage into our instance with `variant.toType(.basis, dest: &value)`

If we now call `init(value:)` we will 
3. Create a second new `Basis()`
4. Call the constructor to copy the storage over again.

PadraigK:
> Another question, you mentioned:
> 
> > I did have to change GodotVariant's init() requirement to be an unwrap so that subclasses of Godot objects could be made to conform to it. This makes the PR a bit more wide-reaching, but it means that it is possible to send a Node object as an argument to a @callable now.
> 
> I could not figure out how Object or Wrapper are made to conform to `GodotVariant` on this patch, did I miss something?
> 
> There are no conformances to it that I could find.

I added these manually in my game project when they were needed. But I'll look at this again as part of re-considering `unwrap()` and see if there's a way to not need to conform in the game project.

PR #196, opened 2023-10-22
Fix #initSwiftExtension failing on empty types

simonomi:
In the [Your First Extension](https://migueldeicaza.github.io/SwiftGodotDocs/tutorials/swiftgodot/your-first-extension) tutorial, section 1, step 6, the documentation suggests the following line:
```swift
#initSwiftExtension(cdecl: "swift_entry_point", types: [])
```
However, this currently causes an error: `Empty collection literal requires an explicit type`, due to the resulting line in the macro expansion `let types = []`.

As the easiest way to fix this, I simply added `: [Wrapped.Type]` into the macro expansion. But while doing this, I thought it could be a nice addition to give the `types` parameter a default value of `[]`. This way, the above macro with no types can be simplified to:
```swift
#initSwiftExtension(cdecl: "swift_entry_point")
```
This has the effect of allowing `types` to be introduced slightly later, which is a fun little bit of progressive disclosure!

And of course, I've added the appropriate test cases. ^-^

### DIFF
--- a/Sources/SwiftGodot/MacroDefs.swift
+++ b/Sources/SwiftGodot/MacroDefs.swift
@@ -60,7 +60,7 @@ public macro Export(_ hint: PropertyHint = .none, _ hintStr: String? = nil) = #e
 /// - Parameter types: The node types that should be registered with Godot.
 @freestanding(declaration, names: named(enterExtension), named(setupExtension))
 public macro initSwiftExtension(cdecl: String,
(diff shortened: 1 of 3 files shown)


PadraigK:
This is cool, it's neat to have the empty case fixed. 

I think I would not default to empty, though because there's almost no point using this macro unless you are sending some types through so it should be encouraged to pass something.

migueldeicaza:

PR #199, opened 2023-10-28
Variant Refactor

PadraigK:
Here is some refactoring in advance of adding back the Signal macro. 

The core idea is to leverage the type system a bit more, which lets us reduce the amount of code we're generating and make it possible to use protocols as generic constraints in the Signal macro stuff (coming soon).

By marking all types that can be stored directly in a native Godot `Variant` as `VariantRepresentable` we can add a single `init(_ variant: some VariantRepresentable)` method to `Variant`, instead of needing a separate initializer for each type. This gives us a contract where we (and the type-checker) know that any `VariantRepresentable` type can travel in and out of a Variant. 

Previously this two-way travel worked, but it was not expressed in a way that the type system could reason about it, so we needed to add things like `toVariant()` and an extra protocol to promise that a given type would implement `toVariant()`.

I'll call out some of the good things here:
* `public required init? (_ from: Variant)` is no longer generated on each object, instead the `VariantStorable` protocol has a default implementation that can initialize any `VariantStorable` from a `Variant`. 
* `public func toVariant () -> Variant` no longer needed and is removed from generation — instead you can just do `Variant(vector)`
* Many initializers are removed from `Variant` as they're now handled by `init(_ variant: some VariantRepresentable)`
* Conformances to `VariantRepresentable` and `VariantStorable` are not code-generated, they're just implemented as extensions. I think this is an improvement because the compiler can tell us about any problems instead of needing to generate the code first. 

And here are some compromises that enable the good things above:
* `init (content: ContentType)` and `init()` are public and required.
* Various ContentType internals are public
~* `VariantCollection < String >`  is no longer possible because String is not a direct 1:1 VariantRepresentable type, so these now generate as `VariantCollection < GString >`. There are still conveniences around String ←→ GString, so I am not sure if this is a problem or not.~ Update: I did some more work so that any `VariantStorable` types (including `String`) can be used anywhere a `VariantRepresentable` can, so this change is no longer necessary.


PadraigK:
API Diffs here: https://gist.github.com/PadraigK/750bfccd51101109689fbe77edaa0a58

PadraigK:
Hold off on reviewing/merging this please, I have a couple more improvements that address the compromises above.

PadraigK:
> Hold off on reviewing/merging this please, I have a couple more improvements that address the compromises above.

Ok, this is in better shape now, it no longer messes up `VariantCollection<String>`

PadraigK:
Updated Diffs: https://gist.github.com/PadraigK/750bfccd51101109689fbe77edaa0a58

Sources/SwiftGodot/Core/VariantRepresentable.swift
```diff
@@ -0,0 +1,193 @@
...
+//  VariantRepresentable.swift
+//
+//
+//  Created by Padraig O Cinneide on 2023-10-22.
+//
+
+/// Types that Variant can hold directly 1:1 with no conversion
+/// Don't conform types that aren't directly representable — instead
+/// use VariantStorable to map to and from VariantRepresentable types.
+public protocol VariantRepresentable: VariantStorable {
```
migueldeicaza, line 11: It would be nice to include a small example here in these API docs.

Sources/SwiftGodot/Core/VariantStorable.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+//
+//  Created by Padraig O Cinneide on 2023-10-22.
+//
+
+@_implementationOnly import GDExtension
+ 
+/// Something that can be wrapped in a Variant, either directly through `VariantRepresentable`
+/// or by converting to a `VariantRepresentable`. 
+public protocol VariantStorable {
```
migueldeicaza, line 12: For the sake of the public API documentation, let us enumerate in this comment examples of each.

migueldeicaza:
My initial take is that I like this patch very much.

I would love if you could review the existing documentation in SwiftGodot/Sources/SwiftGodot/SwiftGodot.docc/Variants.md  as part of this PR, to make sure they do not contradict each other.

migueldeicaza (changes requested):

PadraigK:
> My initial take is that I like this patch very much.
> 
> I would love if you could review the existing documentation in SwiftGodot/Sources/SwiftGodot/SwiftGodot.docc/Variants.md as part of this PR, to make sure they do not contradict each other.

Thanks — I reviewed these and the other two spots you mentioned. Please take another look when you have a chance.

migueldeicaza:

migueldeicaza:
One question: how does this pass a subclass of Wrapper/Object to a Variant?

PR #200, opened 2023-10-29
Fix no-op setter being generated by @Export macro

### LINKED ISSUE #197: Cannot edit exported property
Running on Linux, I cannot edit the value of any `@Export` properties in the Godot editor

```swift
@Godot
class SpinningCube: Node3D {
    @Export(.range, "0,20,")
    public var speed: Double = 1.0;

    @Export
    public var test: Bool = false;

    public override func _ready () {
        let meshRender = MeshInstance3D()
        meshRender.mesh = BoxMesh()
        addChild(node: meshRender)
    }

    public override func _process(delta: Double) {
        if Engine.isEditorHint() { return }
        rotateY(angle: delta * speed)
    }
}
```

The value does show in the inspector, and it *appears* to be editable, but when editing the value, either by moving the slider or typing in a value, the value does not change. Maybe I am doing something wrong?

https://github.com/migueldeicaza/SwiftGodot/assets/36567925/0622d297-9cac-4d0b-ae44-8c5f45cedb8a

Airyzz:
Closes #197 

In all cases where `godotVariants [typeName] != nil`, the resulting code generated by the macro had no operation, and does not set any value. 

For now I have just removed this. I'm not sure if there was a reason for this functionality but this change has fixed the issue I was having. Happy to discuss alternatives if there was a reason for this.

### DIFF
--- a/Sources/SwiftGodotMacroLibrary/MacroExport.swift
+++ b/Sources/SwiftGodotMacroLibrary/MacroExport.swift
@@ -64,7 +64,6 @@ public struct GodotExport: PeerMacro {
         \(varName) = \(typeName)(args [0])!
     """
             }
-            return "func \(name) (args: [Variant]) -> Variant? {\n\treturn nil\n}"
         }
         return "func \(name) (args: [Variant]) -> Variant? {\n\t\(body)\n\treturn nil\n}"
     }


migueldeicaza:

Issue #81, opened 2023-06-05
Not possible to make test cases for SwiftGodot


nvanfleet:
I tried to make some simple tests and realized that it's not possible to even instantiate a node in a TestCase. The crash occurs when the Node tries to instantiate a StringName and then is not capable of doing that because of the not-setup GDExtension code.

IMO one of the main reasons for writing in a compiled language is to be able to write some nice tests for my code. :)

<img width="962" alt="Screenshot 2023-06-05 at 11 44 37 AM" src="https://github.com/migueldeicaza/SwiftGodot/assets/1486399/f7d5d511-1af4-41ab-aea9-a606880eb280">

migueldeicaza:
This is a good observation.

Thinking out loud: There might be an option, using SwiftGodotKit which provides an embeddable version of Godot.  

It is another dependency though and I suspect I need to make it easy to use (it is not currently)

Another option is to look into running tests from Godot.

alicerunsonfedora:
We could likely look into having our own testing suite similar to XCTest and the popular Gut testing suite for Godot (see https://github.com/bitwes/Gut). Though this would likely require either SwiftGodotKit, or an extension of some kind that runs these tests in the editor.

PadraigK:
+1 for this — I wanted to be able to write some tests while hacking on SwiftGodot itself (not in an extension) and there's a similar problem where `gi` is needed for almost anything. 

The SwiftGodotKit idea seems like a good lead to me, but I think there'd need to be some refactoring because it would create a circular dependency right now. 

Is there a a way to just have `libgodot` and a tiny wrapper to get godot launched, without including `SwiftGodot`?

tishin:
This was implemented in #211. I think it's safe to close this issue.

migueldeicaza:
Thanks, closing.

PR #203, opened 2023-10-31
Make it possible to unit test SwiftGodot code

PadraigK:
Posting this PR as a solution to https://github.com/migueldeicaza/SwiftGodot/issues/81 

I copy and pasted from SwiftGodotKit to get this working. It's _pretty_ hacky, but it works for my initial example at least. It adds the `binaryTarget` for `libgodot` but its only linked for tests, not for `SwiftGodot` itself.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -120,7 +120,16 @@ targets.append(contentsOf: [
             "SwiftGodot",
             "ExtensionApi",
             "ExtensionApiJson",
-        ])
+            "libgodot",
+            "GDExtension"
+        ]
+    ),
+    
+    .binaryTarget(
+        name: "libgodot",
+        url: "https://github.com/migueldeicaza/SwiftGodotKit/releases/download/v1.0.1/libgodot.xcframework.zip",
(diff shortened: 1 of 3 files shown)


PadraigK:
Hmm, I'm missing something here — if I add a second test case it crashes. Some global state isn't getting cleared.

migueldeicaza:
I think that it might be necessary to ensure that Godot is only ever initialized once.

I like this approach, even if it duplicates some code - my only concern will be to keep in sync the work in libgodot (which is a fork of Godot) with upstream Godot that this works with.

PadraigK:
@tishin I don't think we should take this as is because it crashes when you add a second test right now. I'm not seeing the test cases in SwiftGodotKit though — do you mean on your own branch? Maybe I'm just not seeing them!

PadraigK:
@tishin If you have a plan though, feel free to take copy my branch and continue from there — we can close this PR, I'll be delighted if we get a way to do unit tests and do not need any credit :D

migueldeicaza:
Closing, as we got the follow up PR merged

PR #205, opened 2023-11-01
Add Signal Macro

PadraigK:
Adds a new freestanding macro that can be used inside a `@Godot` class to register a signal with Godot. 

Signal arguments can be defined and their types will be enforced on new `emit(signal:_argument)` methods.

```swift
@Godot 
class Player: Node2D {
    #signal("game_started")
    #signal("lives_changed", argument: ["new_lives_count": Int.self])

    func startGame() {
       emit(Player.gameStarted)
       emit(Player.livesChanged, 5)
    }
}

class Level: Area2D {
    func _ready() { 
       player.connect(Player.gameStarted, to: self, method: "game_started")
    }
    @Callable func game_started() { 
       GD.print("got game started signal!")
    }
}
```

This all works nicely with the Godot editor too — signals are exposed there and can be connected to GDScripts and well use the argument names provided in the `#signal` invocation. 

---

This PR also fixes an issue where a @Callable functions would not work when called with a GodotObject argument.

### DIFF
--- a/Sources/SimpleExtension/Demo.swift
+++ b/Sources/SimpleExtension/Demo.swift
@@ -23,7 +23,11 @@ class Rigid: RigidBody2D {
 class SwiftSprite: Sprite2D {
     var time_passed: Double = 0
     var count: Int = 0
-
+    
+    #signal("picked_up_item", arguments: ["kind": String.self, "isGroovy": Bool.self, "count": Int.self])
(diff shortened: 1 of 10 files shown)


migueldeicaza:
This is an insta-merge moment.

Thank you so much for your dedication to this effort, it is a gift to all Swift Godot users!

PR #210, opened 2023-11-04
Fix crash when setting incorrect type in editor

### LINKED ISSUE #201: Crash when attempting to set Exported value to incorrect type in editor
When a node has an `@Export` value which takes type of Node or Resource, in the Godot editor, you can try to set the value of this export to a node/resource which does not have the expected type. 

For example

```swift
        @Export
        public var sound: AudioStream?;
```

In the godot editor, I can drag and drop a resource of type `Animation` on to this slot. What should happen here, is the value should just not be set, but it causes an editor crash.

`Could not cast value of type 'SwiftGodot.Animation'  to 'SwiftGodot.AudioStream'`

I was able to track this back to [this line](https://github.com/migueldeicaza/SwiftGodot/blob/a8f5ac0fb9f82eff742bec3ca28f88aacb8266a6/Sources/SwiftGodot/Core/Wrapped.swift#L244)

I also noticed that [here](https://github.com/migueldeicaza/SwiftGodot/blob/a8f5ac0fb9f82eff742bec3ca28f88aacb8266a6/Sources/SwiftGodot/Variant.swift#L325) we are expecting a return type of `T?` where `lookupObject` never actually returns `T?` only `T`. So `asObject` can never actually return `nil` and instead just crashes on error

Airyzz:
Closes #201

### DIFF
--- a/Generator/Generator/ClassGen.swift
+++ b/Generator/Generator/ClassGen.swift
@@ -130,7 +130,7 @@ func generateVirtualProxy (_ p: Printer,
                 if hasSubclasses.contains(cdef.name) {
                     // If the type we are bubbling up has subclasses, we want to create the most
                     // derived type if possible, so we perform the longer lookup
-                    handleResolver = "lookupObject (nativeHandle: resolved_\(i))"
+                    handleResolver = "lookupObject (nativeHandle: resolved_\(i))!"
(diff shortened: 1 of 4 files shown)


migueldeicaza:

PR #211, opened 2023-11-04
Make it possible to unit test SwiftGodot code. Part 2

tishin:
This PR is a continuation of #203. I merged https://github.com/migueldeicaza/SwiftGodotKit/pull/12 into it and made some further developments on the topic:
- It is no longer necessary to have a "main" test. The first executed test method launches godot and runs all the tests inside the runtime.
- It is no longer necessary to manually specify the list of runtime tests. They get automatically discovered.
- It's possible now to execute test methods individually, without executing the entire test target.
- `SwiftGodotTestability` is exposed as a package product, making it possible for users to write tests for their extensions.
- Platform checks added to `libgodot`-dependant functionality, since it's only available on macOS. Cross-platform tests and macOS-specific runtime tests are separated into different test targets.

This pretty much tackles all the issues I described in the original `SwiftGodotKit` PR.


PadraigK:
This is fantastic! I just checked it out and tried adding a few other test cases and it works great. Thanks so much @tishin!

migueldeicaza:
Lovely!   Thank you for this great change!

Issue #214, opened 2023-11-06
Windows macros


Matt-Is-Confused:
Windows got macro support with 5.9.1
Currently macros are completely disabled for windows. 
https://github.com/migueldeicaza/SwiftGodot/blob/main/Sources/SwiftGodot/MacroDefs.swift

PR #215, opened 2023-11-06
Update MacroDefs.swift for windows compat 

Matt-Is-Confused:
Proposed change for #214
Needs to be tested on windows

### DIFF
--- a/Sources/SwiftGodot/MacroDefs.swift
+++ b/Sources/SwiftGodot/MacroDefs.swift
@@ -5,7 +5,7 @@
 //  Created by Miguel de Icaza on 9/25/23.
 //
 
-#if !os(Windows)
+#if !(os(Windows) && swift(>=5.9))
 
 /// Creates the definition for a Swift class to be surfaced to Godot.
 ///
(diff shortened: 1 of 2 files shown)


Sources/SwiftGodot/MacroDefs.swift
```diff
@@ -5,7 +5,7 @@
 //  Created by Miguel de Icaza on 9/25/23.
 //
 
-#if !os(Windows)
+#if !(os(Windows) && swift(>=5.9))
```
tishin, line 8: This will enable macros for Windows with Swift < 5.9

Matt-Is-Confused: good catch I got myself mixed up

Sources/SwiftGodot/MacroDefs.swift
```diff
@@ -5,7 +5,7 @@
 //  Created by Miguel de Icaza on 9/25/23.
 //
 
-#if !(os(Windows) && swift(>=5.9))
+#if !(os(Windows) && swift(<5.9))
```
tishin: This will still leave the macros enabled for swift 5.9 which doesn't have macro support. It should be <=5.9.

Matt-Is-Confused:
This still needs to be tested on a windows install of swift with both 5.9 and 5.9.1 to ensure it doesn't ignore the minor version in the check

migueldeicaza:
Thanks for the contribution!

PR #218, opened 2023-11-07
Update .spi.yml

finestructure:
This updates the `.spi.yml` file to move to hosted documentation at swiftpackageindex.com. As I mentioned on Mastodon, the advantage is that we support versioned docs:

Screenshot 2023-11-07 at 19 10 56

But no worries if you'd rather keep the self-hosted docs and just close the PR if so!

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,3 +1,4 @@
 version: 1
-external_links:
-  documentation: "https://migueldeicaza.github.io/SwiftGodotDocs/documentation/swiftgodot/"
+builder:
+  configs:
+    - documentation_targets: [SwiftGodot, ExtensionApi]
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,4 +1,6 @@
 version: 1
+external_links:
+  documentation: "https://migueldeicaza.github.io/SwiftGodotDocs/documentation/swiftgodot/"
 builder:
   configs:
     - documentation_targets: [SwiftGodot, ExtensionApi]


.spi.yml
```diff
@@ -1,3 +1,4 @@
 version: 1
-external_links:
-  documentation: "https://migueldeicaza.github.io/SwiftGodotDocs/documentation/swiftgodot/"
+builder:
+  configs:
+    - documentation_targets: [SwiftGodot, ExtensionApi]
```
finestructure, line 4: I'm assuming `ExtensionApi` is also to be documented but maybe not? I don't think the current docs include it.

migueldeicaza: This would still keep the documentation that I list here right?

When would that spi.yml documentation be picked up?

finestructure: This would change the documentation link on the SPI package page to the SPI hosted docs (once they're generated) but leave the GH hosted docs untouched.

Screenshot 2023-11-09 at 13 37 30

You make a great point though: we generate the docs for releases as soon as we detect the release (can take up to 4-5h) and for default branch changes within 24h (we coalesce default branch changes over a 24h window).

Now the problem with my initial PR is that by removing the `external_links:` clause I'm removing the doc link until the new docs are in place. So it'd be gone for a good few hours up to 24h.

What we can do instead is add the doc generation config but leave the external link untouched until the docs are actually in place. Then we'll do a second PR to swap out the links.

PR #219, opened 2023-11-08
README: Add Why SwiftGodot and Community sections

superherointj:
My main goal was to document key video on `why SwiftGodot makes sense`:

https://www.youtube.com/watch?v=tzt36EGKEZo

https://media.ccc.de/v/godotcon2023-57866-swift-godot-fixing-the-multi-million-dollar-mistake
Would this link be better?

The lesson I learned from video was:
`No game stutters caused by GC, unlike C#.`

Maybe there is a better way to word this. Feel free to offer suggestion in case it feels wrong.

I moved Slack to community section because I feel this deserves it's own section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,6 @@ to the game engine, or it can be used as an API with SwiftGodotKit
 which embeds Godot as an application that is driven directly from
 Swift.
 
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thanks for the links!

PR #224, opened 2023-11-10
Add convenience inits for modifying after init

Matt-Is-Confused:
Screenshot 2023-11-10 at 10 22 21 AM

Allows for configuration functions to be fed to a class on initialization.


migueldeicaza:
I am unlikely to merge this change for a few reasons:

1. It is not a common Swift idiom.
2. It does not enable functionality that is not currently possible.
3. Most importantly, it introduces a lot of new public methods that count against our limited budget on Windows of 64k public symbols.

You could build a separate library with extensions for your use case that does that though.

PR #227, opened 2023-11-11
Add fallback for builtin constants with missing plain constructors

### LINKED ISSUE #225: Transform3D.identity and other helpful constants
I noticed some useful constants are missing, including `Transform3D.identity`, `Basis.identity`, `Projection.identity`

tishin:
Fixes #225. Though I don't think Extension API's way of describing constant values has anything to do with constructors. It looks more like a memory dump. Anyway, this workaround should patch the existing solution.


migueldeicaza:
Thank you for this patch Tishin!

PR #230, opened 2023-11-12
Add reference count handling for Export macro

### LINKED ISSUE #207: Exported Resource references not loading properly
Steps to reproduce:
- Create a scene, containing a custom node which has an `@Export` resource property
- In the godot editor, set the node's value to an instance of resource
- Close godot editor and reopen
- Upon loading the scene, the node's resource value is cleared

Looking at the .tscn files, after saving and closing, things seem to have been written correctly:
```
[ext_resource type="MyResource" uid="uid://cgrsa4afhcqj4" path="res://resource/value/c_test.tres" id="1_1n7m8"]

[node name="SpinningCube" type="SpinningCube" parent="."]
speed = ExtResource("1_1n7m8")
transform = Transform3D(1, 0, 0, 0, 0.661919, 0.74913, 0, -0.74913, 0.661919, 0, 0, 0)
```

However, after reopening the scene, the file has changed, removing any references to the `1_1n7m8` resource. Im assuming references to `Resource?` are not being loaded properly, and so are defaulting to `nil` and then saving again as `nil`

tishin:
Fixes #207

### DIFF
--- a/Sources/SwiftGodotMacroLibrary/MacroExport.swift
+++ b/Sources/SwiftGodotMacroLibrary/MacroExport.swift
@@ -43,9 +43,12 @@ public struct GodotExport: PeerMacro {
             let optBody = isOptional ? " else { \(varName) = nil }" : ""
             body =
     """
+    	let oldRef = \(varName) as? RefCounted
     	if let res: \(typeName) = args [0].asObject () {
+    		(res as? RefCounted)?.reference()
     		\(varName) = res
     	}\(optBody)
+    	oldRef?.unreference()
     """
         } else {
             if isOptional {


tishin:
Now that I think about it, it's pretty strange that the c++ object gets deallocated by refcount while the corresponding Swift object is still present. I assumed we had something like increasing refcount by 1 in the `RefCounted` init and decreasing it by 1 in the deinit. That was partly the case before https://github.com/migueldeicaza/SwiftGodot/commit/ce3fff5c5efefa1581786d932db94fbd75f09038 as we only had increment on init. Not sure how it works now.

migueldeicaza:
We did have a broken band-aid that always took a reference in Swift to the Reference object, which was incorrect.

Thank you for your patch!  I do not think I had considered this path before.

PR #232, opened 2023-11-15
Add Variant equality operator test

### LINKED ISSUE #159: Variant operator == might be passing the wrong value
I suspect the &ret here needs to be &ret.content:

```
 public static func == (lhs: Variant, rhs: Variant) -> Bool {
...
         gi.variant_evaluate (GDEXTENSION_VARIANT_OP_EQUAL, &lhs.content, &rhs.content, &ret, &valid)
```

tishin:
Closes #159. Reverting to `&ret` makes the test crash, so the issue was fixed in  a2e8854f2e29c8004f85dd99064d22a169e9a4c3


migueldeicaza:
Thank you Mikhail!

Issue #168, opened 2023-10-13
@Godot macro and inheritance 


Matt-Is-Confused:
I attempted to create a base node that I could extend in inheriting nodes
``` Swift
@Godot
class Weapon: Node3D {}

@Godot
class UtilityWeapon: Weapon {}
```
This produces the following error
```
UtilityWeapon.swift:4:1: note: in expansion of macro 'Godot' here
class UtilityWeapon: Weapon {}
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/var/folders/9r/m4y_6wnd0dl9s__40w5yv5n80000gn/T/swift-generated-sources/@__swiftmacro_2XI6Weapon5GodotfMm_.swift:10:12: note: attempt to override property here
static var _initClass: Void = {
```

Removing the macro from the base class does allow it to compile. I am also able to create the nodes in the editor. I am unsure of what functionality is lost by not manually implementing the init code for the base class. Since I don't plan to spawn and use the base class this probably(?) wont be a problem.

migueldeicaza:
Great scenario.  Let me think about how I want to solve this.

Matt-Is-Confused:
A question I just thought of. Does an inheriting class need to re-announce callables and signals from it's baseclass?

migueldeicaza:
It should, yes.

PadraigK:
I was thinking about this — what if the `GodotObject` gains a requirement for a static `initClass` block with a default empty implementation. Then the `init` would be code-genned to always call `let _ = Self.initClass()`

The `@Godot` macro would then just create an `initClass` implementation and never generate initializers. Then it wouldn't be stepping on the subclasses own usage of initializers.

migueldeicaza:
Oh, I like this plan.

EstevanBR:
Just ran into this :)

migueldeicaza:
Kudos to @PadraigK for submitting the fix.

PR #234, opened 2023-11-15
Fixes inheritance when using `@Godot` macro

PadraigK:
Fixes https://github.com/migueldeicaza/SwiftGodot/issues/168

This changes the macro to stop generating `init` methods which were making inheritance a challenge. 

Instead the root class calls an overridable class var with a getter that calls a static method that does the one-time initialization. 

This might look a bit round-about at first glance but works around two limitations in Swift: 
1. Class Initializers are not a supported language thing, so we're approximating them here.
2. class properties can only be calculated properties, not blocks as we would like, so we use a computed class property to call the `static` property that actually does the work. This allows us to support inheritance and ensures that the class initializer is called only once per class.

### DIFF
--- a/Sources/SwiftGodotMacroLibrary/MacroGodot.swift
+++ b/Sources/SwiftGodotMacroLibrary/MacroGodot.swift
@@ -248,7 +248,7 @@ public struct GodotMacro: MemberMacro {
             
             let classInitProperty = DeclSyntax(
             """
-            override class var classInitializer: Void {
+            override open class var classInitializer: Void {
                 let _ = super.classInitializer
                 return _initializeClass
             }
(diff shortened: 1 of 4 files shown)


Sources/SwiftGodotMacroLibrary/MacroGodot.swift
```diff
@@ -265,7 +269,7 @@ public struct GodotMacro: MemberMacro {
                     return stringName
                 }
                 
-                var implementedOverridesDecl = "override class func implementedOverrides() -> [StringName] {\nsuper.implementedOverrides() + [\n"
+                var implementedOverridesDecl = "override open class func implementedOverrides() -> [StringName] {\nsuper.implementedOverrides() + [\n"
```
PadraigK, line 272: This is a bug fix for implementedOverrides; this method needs to be declared open or else classes that import it are not able to override.

migueldeicaza: oops, yes

migueldeicaza:
Thank you so much for taking on this bug @PadraigK !

PR #237, opened 2023-11-16
Fixes issue where Variant parameter to `@Callable` function fails to compile

### LINKED ISSUE #236: Callable with Variant parameter doesn't compile
Let's say I have a method like this:
```swift
    @Callable func setAndLog(property: StringName, value: Variant) {
        GD.print(property)
        self.set(property: property, value: value)
    }
```

This fails to compile with this error:
```
/tmp/swift-generated-sources/@__swiftmacro_9swifttest8InstanceC9setAndLog8CallablefMp_.swift:2:75: error: type 'Variant' has no member 'makeOrUnwrap'
        setAndLog (property: StringName.makeOrUnwrap (args [0])!, value: Variant.makeOrUnwrap (args [1])!)
                                                                         ~~~~~~~ ^~~~~~~~~~~~
Instance.swift:30:15: note: in expansion of macro 'Callable' here
    @Callable func setAndLog(property: StringName, value: Variant) {
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
error: fatalError
```

Generated code:
```
func _mproxy_setAndLog (args: [Variant]) -> Variant? {
	setAndLog (property: StringName.makeOrUnwrap (args [0])!, value: Variant.makeOrUnwrap (args [1])!)
	return nil
}
```

PadraigK:
I think there's an argument that a Variant shouldn't be used as the type here at all. Instead the underlying an object matching the Variant's underlying GType could be used. We _could_ explicitly require `@Callable` params to conform to `VariantStorable` to discourage operating in Variants in your Swift logic and give an error that points the way. I can make a PR to add that restriction if that would be preferable. 

If we just want to make it work for now, this PR should fix #236

### DIFF
--- a/Sources/SwiftGodotMacroLibrary/MacroCallable.swift
+++ b/Sources/SwiftGodotMacroLibrary/MacroCallable.swift
@@ -40,7 +40,13 @@ public struct GodotCallable: PeerMacro {
             if first != "_" {
                 genMethod.append ("\(first): ")
             }
-            genMethod.append ("\(ptype).makeOrUnwrap (args [\(argc)])!")
+            
+            if ptype == "Variant" {
+                genMethod.append ("args [\(argc)]")
+            } else {
+                genMethod.append ("\(ptype).makeOrUnwrap (args [\(argc)])!")
+            }
+            
             argc += 1
         }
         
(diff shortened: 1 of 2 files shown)


migueldeicaza:
I think that there might be scenarios where the user would like to expose the dynamic capabilities of Variant to accept more than one data type, so the approach in this patch is better than limiting users to fully typed APIs.

Thank you for the patch @PadraigK

PR #240, opened 2023-11-18
Change documentation link to SPI

finestructure:
The docs have now been generated for SwiftGodot and can be switched over if you like: https://swiftpackageindex.com/builds/5A228EAD-CFE2-4EB1-AE24-B5F26D0A7FD2

This is just for `main` at the moment until there are future releases with the updated `.spi.yml` file, of course.

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,6 +1,4 @@
 version: 1
-external_links:
-  documentation: "https://migueldeicaza.github.io/SwiftGodotDocs/documentation/swiftgodot/"
 builder:
   configs:
     - documentation_targets: [SwiftGodot, ExtensionApi]


Kyle-Ye:
As I commented in Slack channel. This repo's doc is incomplete now.

So the doc built by SwiftPackageIndex and local Xcode will miss many docs. I'd suggest we revert this change.

eg. See the diff between `https://swiftpackageindex.com/migueldeicaza/swiftgodot/main/documentation/swiftgodot/engine/iseditorhint()` and `https://migueldeicaza.github.io/SwiftGodotDocs/documentation/swiftgodot/engine/iseditorhint()`

I believe the future plan is to remove SwiftGodotDocs repo. Then the local doc build of this repo will be the same as the original SwiftGodotDocs. At that time, we may considering migrating to use SwiftPackageIndex again.

https://github.com/migueldeicaza/SwiftGodot/issues/292

cc @finestructure @migueldeicaza

finestructure:
Just to understand @Kyle-Ye , are we building the docs incorrectly? I spot checked the generated docs when we first created them and they seemed to be the same what was hosted on GitHub.io at the time.

Kyle-Ye:
> Just to understand @Kyle-Ye , are we building the docs incorrectly? I spot checked the generated docs when we first created them and they seemed to be the same what was hosted on GitHub.io at the time.

Some doc info does not exist on https://github.com/migueldeicaza/SwiftGodot. So swiftpackageindex's doc build for SwiftGodot will be incomplete.

See the full context at the Swift on Godot's Slack channel.

image

The merge logic is here

https://github.com/migueldeicaza/SwiftGodot/blob/32089cf8e7d61076264ea190c8d0678438ea3024/Makefile#L13

finestructure:
Ah, I see. That's indeed a process we can't replicate.

Kyle-Ye:
Yes, the current generator relies on Godot source code, so the author may intentionally do not include it in this repository. Instead, a separate repo(SwiftGodotDocs) is used to store the processed products.

Things may change after we move to use JSON file. (Tracked by #292)

migueldeicaza:
I do not intend to delete the SwiftGodotDocs, as I am publishing those.

Additionally, I would like to keep those as the normative, up-to-date docs, but have no problem keeping older versions around for folks that want it.

That said, the issue of requiring a Godot source checkout has now been fixed.

PR #241, opened 2023-11-18
iOS 15 support

jbromberg:
This PR adds support for iOS 15 by using `NSRegularExpression` instead of the new `Regex` that was introduced for iOS 16.

We had to make these changes on our end to support older versions and figured it would be helpful to share.

### DIFF
--- a/Generator/Generator/DocModel.swift
+++ b/Generator/Generator/DocModel.swift
@@ -181,14 +181,10 @@ func loadBuiltinDoc (base: String, name: String) -> DocBuiltinClass? {
     }
 }
 
-@available(macOS 13.0, *)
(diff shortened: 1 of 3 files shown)


migueldeicaza:
These changes have no effect on the iOS platform itself, the code changes are entirely on the generator, which only runs on Mac, Linux, Windows.

I am not sure that this is the right solution to the problem

jbromberg:
Thought so but I couldn't get xcode to stop complaining about using this package in a project to support an older ios version without implementing these changes.

migueldeicaza:
I have an alternative version checked in, can you let me know if that works for you?

PR #243, opened 2023-11-19
docs improvements

yene:
fixes for the steps that stopped me when doing the tutorial for the first time

### DIFF
--- a/Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
+++ b/Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
@@ -65,7 +65,7 @@
         
         @Steps {
             @Step {
-                Create a new Swift file called **PlayerController** in the SimpleRunnerDriver package and import the 
+                Create a new Swift file called **PlayerController.swift** in the `Sources/SimpleRunnerDriver` directory and import the 
                 SwiftGodot library.
                 
                 @Code(name: "PlayerController.swift", file: "PlayerController-starter-init.swift", reset: true)
--- a/Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
+++ b/Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
@@ -144,6 +144,8 @@
             
         @Steps {
             @Step {
+                Download the provided Example project **SimpleRunner-Starter.zip** at the top of the page.
+
                 Create a new `bin` folder in the SimpleRunner Godot project with an empty file 
                 `SimpleRunnerDriver.gdextension`.
                 
--- a/Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
+++ b/Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
@@ -162,7 +162,7 @@
             }
             @Step {
                 Create a `libraries` section and set the path of `macos.debug` to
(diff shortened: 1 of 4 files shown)


Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
```diff
@@ -65,7 +65,7 @@
         
         @Steps {
             @Step {
-                Create a new Swift file called **PlayerController** in the SimpleRunnerDriver package and import the 
+                Create a new Swift file called **PlayerController.swift** in the `Sources/SimpleRunnerDriver` directory and import the
```
alicerunsonfedora, line 68: The tutorial's text clearly states that you're creating a Swift file, I feel the addition of the file extension here is a little redundant?

Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
```diff
@@ -144,6 +144,8 @@
             
         @Steps {
             @Step {
+                Download the provided Example project **SimpleRunner-Starter.zip** at the top of the page.
+
```
alicerunsonfedora, line 148: Most tutorials I've seen that are written with DocC don't explicitly tell the developer to download the package, since it's clearly visible. I think this is a little redundant.

alicerunsonfedora (changes requested):

Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
yene: makes it copy pastable, probably should be in backticks

Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
yene: I created a fresh Godot project only to later realize that I should have used the provided example.

Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
alicerunsonfedora: I don’t think that, in this instance, it is necessary to make it copy/paste-able. While it may be convenient, it doesn’t contribute anything meaningful to the tutorial or the documentation itself.

Sources/SwiftGodot/SwiftGodot.docc/Tutorials/Your First Extension.tutorial
alicerunsonfedora: I feel that there’s already enough context provided in the tutorial to indicate that creating an entirely new project is _not_ the idea:

- First, the abstract of the tutorial states that one will be building off the Simple Runner project, indicating that there is already a starter project to work off of. The abstract is also the first visible thing, since it outlines tutorial length, starter files, etc., and should _not_ be ignored or glossed over.
- Second, the tutorial never calls for making an entirely new Godot project, since the starter already has most of the inner workings together.
- Finally, I feel that adding an instruction to download the starter code when the abstract is already visible may come off as derogatory or demeaning to the developer, belittling them in a way.

alicerunsonfedora:
To reiterate, I am OK with the changes to add the dependencies section. However, I still have concerns over the other changes added that merely provide convenience in the tutorials.

With that in mind, @migueldeicaza if you are fine with these changes, I will not hold out or block.

migueldeicaza:
Let me merge, and we can later fine-tune.

PR #246, opened 2023-11-20
Further fix to support for final classes when using the Godot macro

adamjcampbell:
PR #245 fixed `final class` when there were no method overrides

This PR ensures that `implementedOverrides` also uses the correct access modifier

Without this fix final classes with overrides produce errors in the form of:
Screenshot 2023-11-21 at 7 59 03 am


migueldeicaza:
Thank you for the fix Adam!

PR #251, opened 2023-11-25
Add ExpressibleByStringInterpolation conformance

ulrikdamm:
to StringName, NodePath and GString

Meaning that in addition to creating these from a string literal:
```
let str : StringName = "value"
let node = Node3D("NodeName")
```

You can now also create them from string literals with interpolations:
```
let value = 1
let str : StringName = "value: \(value)"
let node = Node3D("NodeName \(value)")
```


migueldeicaza:
Oh very cool!   Thanks for this contribution!

and I learned something new!

PR #231, opened 2023-11-14
Fix memory leak of framework wrappers

tishin:
Fixes #117

PR #253, opened 2023-11-26
Add memory leak tests

tishin:
Started this as a part of #231, but haven't had a chance to finish it until now.

Commenting out either 
https://github.com/migueldeicaza/SwiftGodot/blob/831e1fefcfaf461554fc5441cfb19df08e485af0/Sources/SwiftGodot/Core/Wrapped.swift#L360
or
https://github.com/migueldeicaza/SwiftGodot/blob/831e1fefcfaf461554fc5441cfb19df08e485af0/Sources/SwiftGodot/Core/Wrapped.swift#L310
triggers the test, so I think it works.


migueldeicaza:
Thank you for these changes!

I love them!

PR #263, opened 2023-12-01
Assert that a class has been registered before use

pcbeard:
This catches the problem of forgetting to call register(type:) on a class marked with the @Godot macro.


migueldeicaza:
This is lovely!   Thank you!

PR #264, opened 2023-12-01
4.2: Remove classes not present in extension_api.json from the known list

tishin:
You're probably not seeing related compilation errors because you have those files cached from 4.1.3 generation.


migueldeicaza:
I had not noticed that, but you are absolutely right, I did not do a clean build.

Thank you for sending this patch!

PR #265, opened 2023-12-02
Add note to windows build instructions

Matt-Is-Confused:
Adds a simple clarification to the windows instructions


migueldeicaza:
I removed all the references to the non-macro world, I think that is better.

PR #272, opened 2023-12-04
Implement codegen for methods using withArgPointers

pcbeard:
This change set refactors methodGen() to generate calls to `gi.object_method_bind_ptrcall` that avoids deeply nested uses of `withUnsafePointer(to:)` in favor of a generated helper function.

        func withArgPointers(
            _ p0: UnsafeRawPointer?, _ p1: UnsafeRawPointer?, _ p2: UnsafeRawPointer?, _ p3: UnsafeRawPointer?, _ p4: UnsafeRawPointer?
        ) {
            var _args: [UnsafeRawPointer?] = [
                p0, p1, p2, p3, p4
            ]
            gi.object_method_bind_ptrcall (Input.method_get_vector, UnsafeMutableRawPointer (mutating: shared.handle), &_args, &_result)
        }

Consider this a work-in-progress. It passes all tests, but needs more cleanup.


Generator/Generator/MethodGen.swift
```diff
@@ -209,7 +341,42 @@ func methodGen (_ p: Printer, method: MethodDefinition, className: String, cdef:
             if args != "" { args += ", "}
             args += "_ arguments: Variant..."
         }
+        // REFACTOR STEP 1:  generate a helper function, a la withUnsafePointers() above, which
+        // combines extracting the parameters into pointers and packing them into the _args array.
+        // We can modularize this by creating functions that generate the return type and return
+        // statements.
+        #if true
```
migueldeicaza, line 348: I am curious, what is the goal for step 2 in this refactoring, and do we need the old code that is on the `#else` part?

Generator/Generator/MethodGen.swift
```diff
@@ -62,6 +62,22 @@ func isRefParameterOptional (className: String, method: String, arg: String) ->
     }
 }
 
+/// TODO: The current code generation for passing parameters is both inefficient, and technically unsafe. We don't need
+/// to use nested invocations of withUnsafePointer to generate pointers to multiple arguments, but we can instead generate
```
migueldeicaza, line 66: Does the TODO here apply to this current PR, or the existing code?

I ask because this meads me to believe that this PR is "technically unsafe", I rather not merge code in that state.   Also curious, what is technical unsafe?

Generator/Generator/MethodGen.swift
```diff
@@ -62,6 +62,22 @@ func isRefParameterOptional (className: String, method: String, arg: String) ->
...
+/// ```
+/// func withUnsafePointers<T1, T2, ReturnType>(
+///     _ p1: UnsafePointer<T1>, _ p2: UnsafePointer<T2>,
+///     _ block: (UnsafePointer<T1>, UnsafePointer<T2>) -> ReturnType
+/// ) -> ReturnType {
+///     block(p1, p2)
+/// }
+/// ```
+/// This reduces the complexity of the generated code, and can be extended to an arbitrary number of parameters.
+///
```
migueldeicaza, line 79: This is lovely!   I learned something today!

Generator/Generator/MethodGen.swift
pcbeard: I was talking about the pre-existing code. Storing the pointers to the arguments in _args, which is in a shallower stack frame.

Generator/Generator/MethodGen.swift
pcbeard: Hopefully we can retire the old code. Just kept it for comparision's sake. Hence the draft PR.

Generator/Generator/MethodGen.swift
pcbeard: I sort of discovered this a few years ago, reasoning that since this is how C APIs are modeled, we could do the same thing in Swift code.

migueldeicaza:
Thank you for this incredible contribution!

Per our discussion, of wanting to keep the #if for a little longer, I merged the patch with the command line to skip the last patch - I believe I got it right, and that there was no functionality lost.   Let me know if I missed something.

pcbeard:
It's easy enough to restore via cherry pick, when you're ready.

Issue #286, opened 2023-12-09
The Slack link on README is no longer valid


Kyle-Ye:
The Slack link on README is no longer valid.

Could anyone update it so that new people can still join the Slack channel?

migueldeicaza:
Updated.   Did not realize this needs to be changed every 30 days, that is kind of obnoxious.

PR #287, opened 2023-12-09
Update slack link to link that does not expire

dvanderleij:
By default when you create an invite link it expires in 30 days. I set this link to never expire


dvanderleij:
see issue #286

PR #293, opened 2023-12-09
Simplify method calls generation

tishin:
Perhaps it would make sense to use a single implementation of `withArgPointers` instead of generating specific ones for each method? It can also be reused to rewrite calls of built-in structs in a similar manner.


migueldeicaza:
Lovely!   Thank you!

PR #294, opened 2023-12-10
Fix exported resource references

### LINKED ISSUE #207: Exported Resource references not loading properly

Airyzz:
Fixes #207 again :p

the value in `\(varName)` is always coming in as nil, and so `reference()` is never being called. I think this was supposed to be adding a reference to `res` anyways. 

looks like incorrect use of `\(varName)` was introduced here: https://github.com/migueldeicaza/SwiftGodot/commit/3f3bf924a533039454af43d3d25b284e0c096858

### DIFF
--- a/Sources/SwiftGodotMacroLibrary/MacroExport.swift
+++ b/Sources/SwiftGodotMacroLibrary/MacroExport.swift
@@ -50,7 +50,7 @@ public struct GodotExport: PeerMacro {
         func dynamicCast<T, U>(_ value: T, as type: U.Type) -> U? { value as? U }
         let oldRef = dynamicCast (\(varName), as: RefCounted.self)
         if let res: \(typeName) = args [0].asObject () {
-            dynamicCast (\(varName), as: RefCounted.self)?.reference()
+            dynamicCast (res, as: RefCounted.self)?.reference()
             \(varName) = res
         }\(optBody)
         oldRef?.unreference()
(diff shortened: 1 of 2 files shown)


Airyzz:
After seeing that build error, I did some more investigating and I think the reason it failed is because `oldRef` is always `nil`, and so it never gets unreferenced. This just didnt show as a problem before because `\(varName)` was never being referenced in the first place and so there was no issue. 

I will look in to get it to unreference properly

migueldeicaza:
Thank you for tracking this down and submitting a fix.

PR #297, opened 2023-12-10
Make @SceneTree path argument optional

### LINKED ISSUE #153: Feature: `@SceneTree` macro should infer path when not provided.
I type this out a lot: `@SceneTree(path: "CollisionShape2D") var collisionShape: CollisionShape2D!`

Since the name of the thing I'm referencing is identical to its type name, it would be nice to have the option to omit the `path` and abbreviate this to: `@SceneTree var collisionShape: CollisionShape2D!`

alicerunsonfedora:
This PR works towards making @SceneTree closer to @BindNode by making the `path` argument optional.

Currently, `@SceneTree` requires an argument for the path:

```swift
class MyNode: Node2D {
	@SceneTree(path: "SomePath/MyOtherNode")
	var myOtherNode: Node2D?
}
```

There may be cases where providing a path is redundant or not desired, however.

This PR changes the macro to allow not providing a path argument. If the path argument is not provided in the macro, it will use the variable declarations identifier text instead of throwing an error.

Consider the following example:

```swift
class MyNode: Node2D {
	@SceneTree var char2D: CharacterBody2D?
}
```

`@SceneTree` would look for a node named `char2D` exactly. This approach makes no assumptions on spacing or case; it is up to the user to ensure their name matches when using the default argument approach. Otherwise, the path argument can be used to explicitly set formatting or pathing.

Fixes #153.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ DerivedData/
 .netrc
 *~
 build-docs.log
+.nova/
--- a/Sources/SwiftGodotMacroLibrary/SceneTreeMacro.swift
+++ b/Sources/SwiftGodotMacroLibrary/SceneTreeMacro.swift
@@ -1,4 +1,4 @@
-// 
+//
 //  SceneTreeMacro.swift
 //  SwiftGodot
 //
@@ -14,7 +14,6 @@ import SwiftSyntaxMacros
 
 public struct SceneTreeMacro: AccessorMacro {
     enum ProviderDiagnostic: String, DiagnosticMessage {
-        case missingPathArgument
         case invalidDeclaration
         case missingTypeAnnotation
         case nonOptionalTypeAnnotation
@@ -23,8 +22,6 @@ public struct SceneTreeMacro: AccessorMacro {
 
         var message: String {
             switch self {
-            case .missingPathArgument:
-                "Missing argument 'path'"
             case .invalidDeclaration:
                 "SceneTree can only be applied to stored properties"
             case .missingTypeAnnotation:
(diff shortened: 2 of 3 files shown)


alicerunsonfedora:
Note: this effectively implements #153.

Kyle-Ye (commented):
After `@SceneTree` add supports to name infer, the only diff between `@SceneTree` and `@BindNode` is that `@BindNode` can provide a nonnull getter.

```swift
@SceneTree var char2D: CharacterBody2D?
@BindNode var char2D: CharacterBody2D
```

But actually we can also achieve the same effect using !
```swift
@SceneTree var char2D: CharacterBody2D!
```

Anyway I think it's time we merge them into 2. So Let's delete the old `@BindNode` implementation in the PR. As for the new name for the macro, `SceneTree` or `BindNode` both is OK for me.

alicerunsonfedora:
> Anyway I think it's time we merge them into one. So Let's delete the old @BindNode implementation in the PR. As for the new name for the macro, SceneTree or BindNode both is OK for me.

The purpose of the PR is to merely provide the default argument, not replace BindNode with SceneTree. This should be done in a future PR with a more vetted process and discussion.

Kyle-Ye (approved):
I'm fine with most of the changes.

One thing to note is that the naming convention for GDScript and Godot Editor seems to use PascalCase for Object name.
eg. Character, LifeBar, ScoreLabel, MobTimer, SpawnLocation

https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html

Should we consider applying such rule to the inferred name?
eg.
```swift
@SceneTree var character: CharacterBody2D?
```
⬇️
```swift
var character: CharacterBody2D? {
    get {
        getNodeOrNull(path: NodePath(stringLiteral: "Character")) as? CharacterBody2D
    }
}
```

alicerunsonfedora:
As @migueldeicaza and I discussed on Slack, we shouldn't make assumptions on how things are named/capitalized:

> I think we should not apply any conversion for that case
> One less thing to explain

https://swiftongodot.slack.com/archives/C05CD0SFZ6X/p1702223708502489

Kyle-Ye:
> As @migueldeicaza and I discussed on Slack, we shouldn't make assumptions on how things are named/capitalized:
> 
> > I think we should not apply any conversion for that case
> > One less thing to explain
> 
> [swiftongodot.slack.com/archives/C05CD0SFZ6X/p1702223708502489](https://swiftongodot.slack.com/archives/C05CD0SFZ6X/p1702223708502489)

Got it.

migueldeicaza:
Would like to have a deprecation warning for the old one, and steer people to the new one, don't want to break people's code needlessly.

migueldeicaza:
Thank you for the change @alicerunsonfedora !

Issue #299, opened 2023-12-11
Cannot build project on windows


Airyzz:
When attempting to build a project on windows, It throws a ton of `lld-link` warnings, and eventually fails:

note, I didn't include all `lld-link` warnings, because there are thousands of them
```
lld-link: warning: D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntaxMacros.build\AbstractSourceLocation.swift.o: locally defined symbol imported: $s11SwiftSyntax11TriviaPieceON (defined in D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntax.build\generated\TriviaPieces.swift.o) [LNK4217]
lld-link: warning: D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntaxMacros.build\AbstractSourceLocation.swift.o: locally defined symbol imported: $s11SwiftSyntax6TriviaV12arrayLiteralAcA0C5PieceOd_tcfC (defined in D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntax.build\Trivia.swift.o) [LNK4217]
lld-link: warning: D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntaxMacros.build\AbstractSourceLocation.swift.o: locally defined symbol imported: $s11SwiftSyntax04ExprB0VMn (defined in D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntax.build\generated\SyntaxBaseNodes.swift.o) [LNK4217]lld-link: warning: D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntaxMacros.build\MacroExpansionContext.swift.o: locally defined symbol imported: $s16SwiftDiagnostics9MessageIDV6domain2idACSS_SStcfC (defined in D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftDiagnostics.build\Message.swift.o) [LNK4217]
lld-link: warning: D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftSyntaxMacros.build\MacroExpansionContext.swift.o: locally defined symbol imported: $s16SwiftDiagnostics0B5ErrorVN (defined in D:\testproject\src\.build\x86_64-unknown-windows-msvc\debug\SwiftDiagnostics.build\Diagnostic.swift.o) [LNK4217]
[...]

PR #300, opened 2023-12-11
Patch to fix windows builds

Airyzz:
A hacky workaround for #299 

Im sure there is probably a more correct way to fix this, but this works for now :>

Also adds a new github action to check windows builds


migueldeicaza:
Oh, I see the patch had exactly what we needed.

Merging.

PR #308, opened 2023-12-14
Create Native Godot functions

nvanfleet:
This is the preliminary addition to replace some of the existing functions from Godot with native ones as well as filtering out some un-needed ones.
- Removes a lot of math functions like sin, cos, tan from being generated. These are unecessary because Swift already has those
- Removes lerp and inverse_lerp and implements a native version
- Removes snapped and implements a native version
- Removes angle calculations and implements a native one

### DIFF
--- a/Generator/Generator/ClassGen.swift
+++ b/Generator/Generator/ClassGen.swift
@@ -202,7 +202,7 @@ func generateVirtualProxy (_ p: Printer,
 }
 
 // Dictioanry of Godot Type Name to array of method names that can get a @discardableResult
-var discardableResultList: [String: Set<String>] = [
+let discardableResultList: [String: Set<String>] = [
     "Object": ["emit_signal"],
     "GArray": ["append"],
     "PackedByteArray": ["append"],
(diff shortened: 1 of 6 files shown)


Generator/Generator/ClassGen.swift
```diff
@@ -202,7 +202,7 @@ func generateVirtualProxy (_ p: Printer,
 }
 
 // Dictioanry of Godot Type Name to array of method names that can get a @discardableResult
-var discardableResultList: [String: Set<String>] = [
```
nvanfleet, line 205: I couldn't really resist changing this

Generator/Generator/ClassGen.swift
```diff
@@ -219,9 +219,24 @@ var discardableResultList: [String: Set<String>] = [
     "RefCounted": ["reference", "unreference"]
 ]
 
-func generateMethod (_ p: Printer, method: MethodDefinition) {
```
nvanfleet, line 222: This seems this was never used?

migueldeicaza:
I love these changes - Will merge to get the ball rolling, but would love to have tests associated with these, so we know that we are not regressing things for folks.

jbromberg:
FYI there's a typo here on the `radiansToDegreess` computed property

migueldeicaza:
Thanks, fixed!

PR #309, opened 2023-12-16
Fix Linear Interpolation

nvanfleet:
Long story short my original linear interpolation was bugged. I had tested things but then made a few mistakes trying to arrange how the extension worked and caused an issue. So here is the fix with a cleanup as well as tests for snapping and linear interpolation.


migueldeicaza:
Very nice!  Love the tests!

PR #325, opened 2024-01-03
Fix StringName argument generation

### LINKED ISSUE #324: Incorrect `StringName` default values in method arguments
Some of the methods have `StringName` default arguments that don't match Godot documentation, e.g. in `AnimatedSprite2D`:
```swift
public final func play (name: StringName = StringName ("dv"), customSpeed: Double = 1.0, fromEnd: Bool = false)
public final func playBackwards (name: StringName = StringName ("dv"))
```

It should be an empty string according to the [docs](https://docs.godotengine.org/en/stable/classes/class_animatedsprite2d.html#class-animatedsprite2d-method-play)

tishin:
Fixes #324

### DIFF
--- a/Generator/Generator/Arguments.swift
+++ b/Generator/Generator/Arguments.swift
@@ -29,7 +29,7 @@ func getArgumentDeclaration (_ argument: JGodotArgument, eliminate: String, kind
     let optNeedInOut = ""
     
     var def: String = ""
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thanks for spotting the bug and providing a fix tishin!

PR #326, opened 2024-01-03
Add @Callable macro explainer in the documentation.

jtodaone:
I had a hard time figuring out why my Callable made with the macro was making the game freeze, turns out you have to call `<callable>.call()` method instead of just using `<callable>()`

I don't know if this behavior is a bug or not, but I think it might be worth it to include GDScript snippet to let the users know how to do this.


migueldeicaza:
Thank you for this documentation update!

PR #332, opened 2024-01-05
Fix Quaternion default init

### LINKED ISSUE #331: `Quaternion()` default init does not match its GDScript counterpart
Despite the [docs](https://docs.godotengine.org/en/stable/classes/class_quaternion.html#class-quaternion-constructor-quaternion)  for default constructor saying `Constructs a default-initialized quaternion with all components set to 0.`, GDScript actually initializes quaternion with `identity` values (0, 0, 0, 1), which makes more sense.

SwiftGodot, on the other hand, constructs (0, 0, 0, 0) in `init()`

tishin:
Fixes #331


migueldeicaza:
Nicely spotted, thank you once again!

PR #334, opened 2024-01-07
Add engine tests for AABB, Basis, Plane and Quaternion

tishin:
These are just C++ tests from Godot's repository converted into Swift. It's a relatively cheap way for us to get some test coverage, and to make sure things work as the engine devs intended.


migueldeicaza:
Beautiful and outstanding!   I am deeply grateful for this.

PR #337, opened 2024-01-08
Specify Int size for generated enums

### LINKED ISSUE #336: Global scope enums `rawValue` mismatch
[Global scope enums](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enumerations) are currently generated with `Int` as the `RawValue` type. 
There are a number of methods that don't use those enum types explicitly, but accept the underlying primitive types, e.g. [Basis.fromEuler](https://docs.godotengine.org/en/stable/classes/class_basis.html#class-basis-method-from-euler) or [Rect2.growSide](https://docs.godotengine.org/en/stable/classes/class_rect2.html#class-rect2-method-grow-side). And those methods are generated with `Int64` arguments. This results in a type mismatch not allowing direct use of the raw values:
```swift
Basis.fromEuler(euler: Vector3.zero, order: Int64(EulerOrder.xyz.rawValue))
Rect2().growSide(side: Int64(Side.top.rawValue), amount: 0)
```
Those methods that do accept enum values like `Input.isKeyPressed(keycode: Key.space)` or `Image().rotate90(direction: ClockDirection.clockwise)` also convert raw values to `Int64` under the hood.

tishin:
Fixes #336

### DIFF
--- a/Generator/Generator/Enums.swift
+++ b/Generator/Generator/Enums.swift
@@ -83,7 +83,7 @@ func generateEnums (_ p: Printer, cdef: JClassInfo?, values: [JGodotGlobalEnumEl
             p.indent += 1
             enumDefName = String (enumDefName.dropFirst("Variant.".count))
         }
-        p ("public enum \(getGodotType (SimpleType (type: enumDefName))): Int") {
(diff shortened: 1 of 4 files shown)


migueldeicaza:
Thank you for tracking this down!

PR #340, opened 2024-01-09
Omit first argument name in built-in struct methods when possible

tishin:
There's already similar logic for general classes, but not for built-ins.


migueldeicaza:
Oh this is delightful!

Issue #327, opened 2024-01-04
Callable Macro does not support throwing / async function.


jtodaone:
It seems like currently Callable Macro does not recognize async functions or throwing functions.

Example code:

```swift
@Callable func test() async throws {
    
}
```

Generated code:

```swift
func _mproxy_test (args: [Variant]) -> Variant? {
    test ()
    return nil
}
```

It's very much of a half-baked solution but maybe it should \*at least\* be like this:

```swift
func _mproxy_test (args: [Variant]) -> Variant? {
    Task {
        try? await test ()
    }
    return nil
}
```

Ideally, it should be connected to Godot's native asynchronous programming rules and error handling rules, but I am not well-versed in that.

+ Currently, as a workaround, you can get around this by wrapping it in a separate Callable function:

```swift
@Callable func test() {
    Task {
        try? await _test()
    }
}

func _test() async throws {
    
}
```

jtodaone:
I tried to implement this, but then I realized sometimes async functions will need to return values and my proposed quick fix cannot do that. Is there a way to send something like a `Future<>` or `Promise<>` to Godot?

solgar:
Given that limitation I don't understand how timers can be awaited. Timers have `emitted` property, but there is no actual way of awaiting it.

Code from `Dodge The Creeps` which is example game from official Godot tutorial has following code in `hud.gd` file:
```gdscript
func show_message(text):
	$Message.text = text
	$Message.show()
	$MessageTimer.start()

func show_game_over():
	show_message("Game Over")
	# Wait until the MessageTimer has counted down.
	await $MessageTimer.timeout

	$Message.text = "Dodge the\nCreeps!"
	$Message.show()
	# Make a one-shot timer and wait for it to finish.
	await get_tree().create_timer(1.0).timeout
	$StartButton.show()
```
Which is called from `main.gd`:
```gdscript
func game_over():
	$ScoreTimer.stop()
	$MobTimer.stop()
	$HUD.show_game_over()
	$Music.stop()
	$DeathSound.play()
```
I tried to adapt it as `hud.swift`:
```swift
  func showMessage(_ msg: String) {
    messageLabel.text = msg
    messageLabel.show()
    messageTimer.start()
  }

  func showGameOver() async {
    showMessage("Game Over")
    await messageTimer.timeout.emitted
      
    messageLabel.text = "Dodge the\nCreeps!"
    messageLabel.show()

[...]

PR #341, opened 2024-01-09
Throw an error when async or throwing function have @Callable macro

jtodaone:
Following the discussion on #327, I have implemented a check for async/throwing functions since currently it just does not compile and does not give an explanation on why.

This PR makes the code below throw an error before compile-time.

```swift
@Callable func test() async throws { // ERROR: @Callable does not support asynchronous or throwing functions
    print("test")
}
```


migueldeicaza:
Thank you, thank you!   This is a great usability touch!

PR #343, opened 2024-01-11
Add engine tests for Rect2i and Rect2

tishin:


migueldeicaza:
Glorious, thank you so much!

PR #351, opened 2024-01-14
Prevent SwiftGodot subclasses from executing in editor by default

### LINKED ISSUE #347: Extension subclasses run in editor
In GDScript scripts don't run in editor by default, there's a special annotation for that: [`@tool`](https://docs.godotengine.org/en/stable/tutorials/plugins/running_code_in_the_editor.html)

That's not the case for GDExtension. I assume, it's by design that bindings have to implement game and editor logic separation.
For example, in Rust bindings subclasses don't run as tools by default, but [can be declared as such](https://godot-rust.github.io/docs/gdext/master/godot/prelude/derive.GodotClass.html#running-code-in-the-editor). 

Currently Swift subclasses run in editor, and the only way I know of to work around that is to use `Engine.isEditorHint()` checks in gameplay code. I think it's counterintuitive, and can lead to a lot of confusion (#77, for example).

tishin:
Fixes #347. Classes can be annotated as `@Godot(.tool)` to make them run in editor akin to `@tool` scripts.

### DIFF
--- a/Sources/SwiftGodot/MacroDefs.swift
+++ b/Sources/SwiftGodot/MacroDefs.swift
@@ -14,9 +14,16 @@
 /// any `@Export` and `@Callable` methods for the class effectively surfacing properties and
 /// methods to godot
 ///
+/// - Parameter behavior: using `.tool` value makes the overridden methods like `_ready` or
+/// `_process` run in editor, making the class work like `@tool` annotated script in GDScript
+///
 @attached(member,
           names: named (_initializeClass), named(classInitializer), named (implementedOverrides))
-public macro Godot() = #externalMacro(module: "SwiftGodotMacroLibrary", type: "GodotMacro")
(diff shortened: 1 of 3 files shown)


migueldeicaza:
So this does not rely on upcoming features in GDExtension?

I could swear I saw a proposal for something new coming up.

tishin:
Yes, it's just a band-aid on SwiftGodot side, similar to what Rust bindings have to adjust default GDExtension behavior.
There's a PR open for Godot 4.3 addressing the issue from GDExtension side, but I'm not sure what state it's in.

migueldeicaza:
I found the PR.

We can revisit this when 4.3 is out:

https://github.com/godotengine/godot/pull/82554

migueldeicaza:
Thank you for this change and looking into what Rust did.

cplr:
I believe the example script in the main repo README needs to be updated to feature this annotation, as without it, you do not see the added cube mesh in the editor.

PR #352, opened 2024-01-16
Reuse collection conformance generation for GArray

tishin:
This will hook `GArray` to the generated indexing implementation already used in packed arrays.


migueldeicaza:
This is lovely!

Thank you!

PR #359, opened 2024-01-20
Add unary minus operators

### LINKED ISSUE #356: Unary minus operators aren't being generated
In GDScript it's possible to use [unary minus](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-operator-unminus) operators for negation, e.g. `print(-Vector3(1, 2, 3))` will output `(-1, -2, -3)`.
Currently we skip all unary operators during generation. Not sure if we need "unary plus" or "not" operators though.

tishin:
Fixes #356


migueldeicaza:
Lovely, thank you very much!

PR #364, opened 2024-01-22
Fix conflicting variable name in export macro

### LINKED ISSUE #363: Cannot export a variable with name "value"
I am updating my project to the latest version of SwiftGodot, and have run in to this issue. 

In my project, I have some custom resources which simply exports some value:

```swift
@Godot
class IntResource: Resource {
    public var hasChanged = false

    public var innerValue: Int = 0

    @Export
    public var value: Int {
        get {
            if hasChanged {
                return innerValue
            }
            return defaultValue
        }
    }

    @Export
    public var defaultValue: Int = 0
}
```

This is no longer possible, as the code generation for this export redefined 'value' to be a `let`:
```swift
func _mproxy_set_value (args: [Variant]) -> Variant? {
	guard let arg = args.first else {
		return nil
	}
	if let value = Int (arg) {
		value = value
	} else {
		GD.printErr ("Unable to set `value` value: ", arg)
	}
	return nil
}
```

Airyzz:
Fixes #363

### DIFF
--- a/Sources/SwiftGodot/Core/Wrapped.swift
+++ b/Sources/SwiftGodot/Core/Wrapped.swift
@@ -312,8 +312,9 @@ func lookupObject<T:GodotObject> (nativeHandle: UnsafeRawPointer) -> T? {
         } else {
             print ("Found a custom type for \(className) but the constructor failed to return an instance of it as a \(T.self)")
         }
-    } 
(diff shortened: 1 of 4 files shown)


migueldeicaza:
Thank you!

Issue #330, opened 2024-01-05
Feature Request: Way to disable SwiftGodot's calls to print()


Airyzz:
I am currently working on a project using the godot-jolt physics engine. Because this is also implemented via GDExtension, SwiftGodot is unable to get the class `JoltPhysicsDirectBodyState3D` and falls back to `PhysicsDirectBodyState3D`. This is fine, and it still functions as expected, except for that my terminal gets filled with: `Could not find class JoltPhysicsDirectBodyState3D, fallback to creating a PhysicsDirectBodyState3D` every frame, making it frustrating to read the logs that I actually care about (my own calls to print)

Additionally, every time I close the game my terminal is filled with thousands of messages like: 

```
SWIFT: Removed from our live Objects with key 0x000000000c29a080, removed: <Object#0>
SWIFT: Removed object from our live SubType list (type was: <RollbackCollisionObject3D#51103401628>
SWIFT: instanceBindingFree token=Optional(0x000000000682ef20) instance=Optional(0x000000000b2ca840) binding=Optional(0x000000000b444c30)
SWIFT: Removed object from our live SubType list (type was: <ShootableRigidBody3D#51069847194>
SWIFT: instanceBindingFree token=Optional(0x000000000682ef20) instance=Optional(0x000000000b299040) binding=Optional(0x000000000a0e88f0)
SWIFT: Removed object from our live SubType list (type was: <GameModeDebug#50516199041>
SWIFT: instanceBindingFree token=Optional(0x000000000682ef20) instance=Optional(0x000000000b4e0f90) binding=Optional(0x000000000b27ffd0)
SWIFT: Removed object from our live SubType list (type was: <EventSceneLoader#42748347720>
SWIFT: instanceBindingFree token=Optional(0x000000000682ef20) instance=Optional(0x000000000b1c2d80) binding=Optional(0x000000000b6c38f0)
SWIFT: Removed object from our live SubType list (type was: <EventSceneLoader#42731570501>
SWIFT: instanceBindingFree token=Optional(0x000000000682ef20) instance=Optional(0x000000000b1c2950) binding=Optional(0x000000000b6be5d0)
SWIFT: Removed object from our live SubType list (type was: <DebugLine#29712450884>
SWIFT: instanceBindingFree token=Optional(0x000000000682ef20) instance=Optional(0x000000000b034570) binding=Optional(0x000000000b032100)
[...]

PR #365, opened 2024-01-22
Remove annoying prints

Airyzz:
Can we please get rid of these? They are flooding my console output with information that I dont care about.

Relates to #330


migueldeicaza:
Done!

PR #368, opened 2024-01-23
Fix arithmetic overflow crash in unary minus operators

tishin:
Follow-up to #359. Godot does not crash on arithmetic overflows, so neither should we.


migueldeicaza:
I wonder if we should provide alternative operators that do not crash - so that we can preserve the Swift behavior of trapping on overflows.

tishin:
I'd rather have alternatives that do crash. I don't think making native Godot interfaces crash under conditions when they're not supposed to crash is a good idea. That combined with Godot not having a proper crash reporting system is a nightmare fuel for me. Bindings users shouldn't wonder whether an operator or a method has a native non-crashing implementation or it's a Swifty crashing overload.

migueldeicaza:
Makes sense.

PR #372, opened 2024-01-29
Make default Color init match GDScript

tishin:
In GDScript `Color()` constructs a non-transparent black: `(0, 0, 0, 1)`


migueldeicaza:
Thank you!   Another great find

PR #375, opened 2024-01-30
Add RefCounted.unref() to manage lifetime of RefCounted objects

pcbeard:
This calls existing extension function object_destroy() when the object's reference count goes to 0.

It is a partial solution to this issue https://github.com/migueldeicaza/SwiftGodotKit/issues/25.


pcbeard:
To truly fix this completely we might want to define a property wrapper for owned references to `RefCounted` objects. Currently our RefCounted class makes no ownership assumptions.

migueldeicaza:
Related fix, we currently leak variants, which is only a problem with those that come with juicy payloads.

I have a patch for that, that I have been testing.   This no longer crashes for me:

https://gist.github.com/migueldeicaza/652c9665459cf05a3cad856ad91d11c8

migueldeicaza:
I have manually merged the 'unref' bits.

I would love to merge some of the overall build-system changes to build on Linux, independently of the actual test suite changes.   I think those would be nice to have while we sort out the rest.

migueldeicaza:
A more complete and comprehensive fix is now part of 0.45 (about to be released).

migueldeicaza:
Reopening, as this has other interesting bits

migueldeicaza:
Closing, as we have now a fully revamped one here: https://github.com/migueldeicaza/SwiftGodot/pull/663

PR #376, opened 2024-01-30
Add HSV Color properties

tishin:
Those are described as properties in Godot docs: [h](https://docs.godotengine.org/en/stable/classes/class_color.html#class-color-property-h), [s](https://docs.godotengine.org/en/stable/classes/class_color.html#class-color-property-s), [v](https://docs.godotengine.org/en/stable/classes/class_color.html#class-color-property-v). They are even exposed as properties in GDExtension, but, I assume, since they are computed properties and not struct members, we just skip them during generation.

This implementation is based on Godot's C++ implementation.


migueldeicaza:
Once again, I am incredibly thankful for your contribution.

Cheers!

PR #382, opened 2024-02-01
Mark pushBack method result as discardable in packed arrays

tishin:
`append` and `push_back` are the same


migueldeicaza:
Agreed. Thanks!

PR #385, opened 2024-02-05
Add local scope subtypes registration interface for tests

tishin:
This replaces global type registration list in tests with local scope lists, allowing each `GodotTestCase` to register its Godot subtypes before test execution and unregister them after. Should give an option to use private access for subtypes and not to worry about name intersections with subtypes from other tests.


migueldeicaza:
Oh this is lovely!

PR #387, opened 2024-02-06
Make StringName and GString Hashable

### LINKED ISSUE #344: Make StringName hashable
I believe this should be possible, but need to validate.

### LINKED ISSUE #379: Make StringName Hashable
```
    public func hash(into hasher: inout Hasher) {
        hasher.combine(self.hash ())
    }
```

tishin:
Closes #344, closes #379. 

It's probably possible to make it match Swift string hashing:
```swift
extension String: Hashable {
  /// Hashes the essential components of this value by feeding them into the
  /// given hasher.
  ///
  /// - Parameter hasher: The hasher to use when combining the components
  ///   of this instance.
  public func hash(into hasher: inout Hasher) {
    if _fastPath(self._guts.isNFCFastUTF8) {
      self._guts.withFastUTF8 {
        hasher.combine(bytes: UnsafeRawBufferPointer($0))
      }
      hasher.combine(0xFF as UInt8) // terminator
    } else {
      _gutsSlice._normalizedHash(into: &hasher)
    }
  }
}
```
by using `toUtf8Buffer`, but I don't think there's an effective way to get actual byte array from `PackedByteArray`.

### DIFF
--- a/Sources/SwiftGodot/Core/StringExtensions.swift
+++ b/Sources/SwiftGodot/Core/StringExtensions.swift
@@ -7,7 +7,7 @@
 
 @_implementationOnly import GDExtension
 
-extension StringName: CustomStringConvertible {
+extension StringName: CustomStringConvertible, Hashable {
     /// Creates a StringName from a Swift String.Substring
     public convenience init (_ from: String.SubSequence) {
         self.init (from: String (from))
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Oh this is delightful!   Thank you @tishin !

PR #389, opened 2024-02-09
Fix Variant ret

nvanfleet:
I was getting this compile error

```
/Users/nvanfleet/src/PlantQuest/build/checkouts/SwiftGodot/Sources/SwiftGodot/Variant.swift:185:21: error: use of local variable 'ret' before its declaration
        if let rc = ret as? RefCounted {
                    ^
/Users/nvanfleet/src/PlantQuest/build/checkouts/SwiftGodot/Sources/SwiftGodot/Variant.swift:189:13: note: 'ret' declared here
        let ret: T? = lookupObject(nativeHandle: value)
            ^
error: fatalError
```

Looks like the ret should be declared before it is used.

cc @migueldeicaza


migueldeicaza:
Thank you!

PR #391, opened 2024-02-11
Fix destruction of externally managed variants

### LINKED ISSUE #390: `VariantCollection` crashes after Variant leak fix
That's the reason CI tests are failing. Simplified example:
```swift
func testClipPolygons () {
    let a = PackedVector2Array ()
    let b = PackedVector2Array ()
    var r: VariantCollection<PackedVector2Array>
    
    a.pushBack (value: Vector2 (x: 225, y: 180))
    a.pushBack (value: Vector2 (x: 160, y: 230))
    a.pushBack (value: Vector2 (x: 20, y: 212))
    a.pushBack (value: Vector2 (x: 50, y: 115))
    
    b.pushBack (value: Vector2 (x: 150, y: 200))
    b.pushBack (value: Vector2 (x: 65, y: 190))
    b.pushBack (value: Vector2 (x: 80, y: 140))

    r = Geometry2D.clipPolygons (polygonA: a, polygonB: b)
    XCTAssertEqual (r.size (), 2, "Polygon 'a' completely overlaps polygon 'b'. This should result in 2 clipped polygons.")
    XCTAssertEqual (r [0].size (), 4, "The resulting clipped polygon should have 4 vertices.")
    XCTAssertEqual (r [0].size (), 4, "The resulting clipped polygon should have 4 vertices.")
}
```

The first `r[0].size()` check works fine, the second one crashes.

tishin:
Fixes #390. Not sure if it's the correct solution, but the way I see it, Variant shouldn't destroy `content` which wasn't initialized for that variant and is managed somewhere else.

### DIFF
--- a/Sources/SwiftGodot/Variant.swift
+++ b/Sources/SwiftGodot/Variant.swift
@@ -67,12 +67,18 @@ public class Variant: Hashable, Equatable, CustomDebugStringConvertible {
     var content: ContentType = (0, 0, 0)
     static var zero: ContentType = (0, 0, 0)
     
+    /// Underlying Godot Variant should be destroyed on Swift Variant deinit
+    var destroyOnDeinit: Bool = true
+    
     /// Initializes from the raw contents of another Variant
     init (fromContent: ContentType) {
         content = fromContent
+        // Content deallocation is managed externally
+        destroyOnDeinit = false
     }
     
     deinit {
+        guard destroyOnDeinit else { return }
         if experimentalDisableVariantUnref { return }
         gi.variant_destroy (&content)
     }


migueldeicaza:
This is roughly in the ballpark of what needs to be done.

But the problem is that this constructor can be used for legitimate ownership creation, so I would like instead to have an explicit call to the `variant_copy` methods on those cases.   My issue is that the documentation for this is non-existent, and the ownership model that we currently have is an approximation to reality based on existing patterns and idioms.

So I would love to find what triggered this crash for you, so I can alter that codepath to call the Variant constructor that makes a copy.

migueldeicaza:
Oh!   I see that there is a bug linked to this!

Let me take a look.

migueldeicaza:
Ok, I took a slightly different approach.

The tests work (even if I can not run them on Xcode anymore)

PR #394, opened 2024-02-13
Restructure VariantRepresentable conformances

### LINKED ISSUE #392: `VariantCollection` crashes when working with `ContentTypeStoring` types
A simple `append` is enough to reproduce the crash:
```swift
let c = VariantCollection<PackedVector2Array>()
c.append(value: PackedVector2Array())
```

This probably happens due to incorrect usage of `Variant.init`:
```swift
init<T: VariantRepresentable>(representable value: T) where T: ContentTypeStoring
init<T: VariantRepresentable>(representable value: T)
```
`VariantCollection<PackedVector2Array>` invokes the latter while `PackedVector2Array` should go through the former.

tishin:
Fixes #392. Forces `VariantRepresentable` conformants to provide their bounding types and values:
-  `self` for structs/scalars
- `contentType` for ex-`ContentTypeStoring` types
- `handle` for `Object`

This should resolve any possible issues related to `VariantRepresentable` abstractions not accounting for `ContentTypeStoring`


migueldeicaza:
This is beautiful, much better than the manual band-aid I applied.

PR #402, opened 2024-02-17
Make SceneTree macro path optional

### LINKED ISSUE #401: Omitting `@SceneTree.path` breaks compilation
For some reason example from #297 does not compile anymore:
```swift
@Godot
class MyNode: Node2D {
    @SceneTree var char2D: CharacterBody2D? // Type of expression is ambiguous without a type annotation
}
```
Looks like a regression.

tishin:
Fixes #401. Not sure how it managed to work previously, `path` wasn't optional and didn't have a default value in the original PR. Maybe Xcode/Swift updates broke it, I don't know.


migueldeicaza:
Thank you Mikhail!

PR #404, opened 2024-02-19
Make async interface of signals connect on main thread

### LINKED ISSUE #403: Unable to use `await` for `Timer.timeout` signal
A simplified node example:

```swift 
@Godot
class TimerNode: Node {
    
    let timer = Timer()
    
    override func _ready() {
        addChild(node: timer)
        
        timer.start()
        GD.print("started")
        
        Task {
            await timer.timeout.emitted
            GD.print("emitted")
        }
    }
    
}
```

Awaiting for timeout throws an error:
```
ERROR: Caller thread can't call this function in this node (/root/Timer/@Timer@2). Use call_deferred() or call_thread_group() instead.
at: connect (scene/main/node.cpp:3706)
Warning, error connecting to signal timeout: errInvalidParameter
```
And the second print does not happen.

tishin:
Fixes #403

### DIFF
--- a/Sources/SwiftGodot/Core/SignalSupport.swift
+++ b/Sources/SwiftGodot/Core/SignalSupport.swift
@@ -135,6 +135,7 @@ public class SimpleSignal {
     }
     
     /// You can await this property to wait for the signal to be emitted once
+    @MainActor
     public var emitted: Void {
         get async {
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thank you for this fix!   I hope we can still raise signals intended for other threads

PR #406, opened 2024-02-21
Replace NSRegularExpressions with native Swift Regexes

Zollerboy1:
I don't think anyone should use `NSRegularExpression` in Swift anymore, so I replaced the few places where it is used with the native Swift Regexes that are available since Swift 5.7.


migueldeicaza:
That is what we used initially, but they are not supported on Windows, hence why we went that route.

Once they are supported on Windows, we can change them, but until then, it prevents SwiftGodot from being usable by our Windows users.

PR #420, opened 2024-03-11
Bump dependencies versions

tishin:


migueldeicaza:
One concern about bumping this value is that we have some scary warnings that I think we should fix before we bump:

> 'as' is deprecated: This cast will always fail
> 'as' is deprecated: This cast will always succeed

Those do look worrisome.

migueldeicaza:
I found the release notes here:

https://github.com/apple/swift-syntax/releases

PR #423, opened 2024-03-12
Fix build command execution on Windows using Swift 5.10

### LINKED ISSUE #418: Windows: Unable to build SwiftGodot using Swift 5.10 toolchain
Building SwiftGodot on Windows using Swift 5.10 fails during generator plugin execution:
```
error: failed: PrebuildCommand(configuration: SPMBuildCore.BuildToolPluginInvocationResult.CommandConfiguration(displayName: Optional("Generating Swift API from /Projects/SwiftGodot/Sources/ExtensionApi/extension_api.json to /Projects/SwiftGodot/.build/plugins/outputs/swiftgodot/SwiftGodot/CodeGeneratorPlugin/GeneratedSources"), executable: <AbsolutePath:"\Projects\SwiftGodot\.build\plugins\tools\debug\Generator.exe">, arguments: ["/Projects/SwiftGodot/Sources/ExtensionApi/extension_api.json", "/Projects/SwiftGodot/.build/plugins/outputs/swiftgodot/SwiftGodot/CodeGeneratorPlugin/GeneratedSources", "/Projects/SwiftGodot/doc", "--singlefile"], environment: [:], workingDirectory: nil), outputFilesDirectory: <AbsolutePath:"\Projects\SwiftGodot\.build\plugins\outputs\swiftgodot\SwiftGodot\CodeGeneratorPlugin\GeneratedSources">)

Generator/main.swift:30: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"
Current stack trace:
0    (null)                             0x00007ffac5964ae0 swift_stdlib_reportFatalErrorInFile + 132
```

tishin:
Fixes #418 

We have two build commands here:
1. `prebuildCommand` which works as a source generator on Windows, but shouldn't work by design (on macOS it throws `a prebuild command cannot use executables built from source, including executable target 'Generator'`). On Swift 5.9 this command also added sources from `outputFilesDirectory` to the build, which does not happen on Swift 5.10. Not sure if we should report this, as our use case is wrong to begin with.
2. `buildCommand` which does not generate sources on Windows (the `.exe` never gets called), but works as a source provider through `outputFiles`. I have already reported the issue at https://github.com/apple/swift-package-manager/issues/6962

So combining the two broken things makes the build work again on Swift 5.10 toolchain. ~This shouldn't affect 5.9 builds, as `buildCommand` does not add any actual workload to the build.~
I wish there was a less hacky solution.

### DIFF
--- a/Plugins/CodeGeneratorPlugin/plugin.swift
+++ b/Plugins/CodeGeneratorPlugin/plugin.swift
@@ -37,12 +37,15 @@ import PackagePlugin
         outputFiles.append (contentsOf: known.map { genSourcesDir.appending(["generated", $0])})
         #endif
 
+        // For Windows with Swift 5.10 both prebuildCommand and buildCommand are needed
+        #if !os(Windows) || swift(>=5.10)
         commands.append(Command.buildCommand(
             displayName: "Generating Swift API from \(api) to \(genSourcesDir)",
             executable: generator,
             arguments: arguments,
             inputFiles: [api],
             outputFiles: outputFiles))
+        #endif
         
         return commands
     }
--- a/Plugins/CodeGeneratorPlugin/plugin.swift
+++ b/Plugins/CodeGeneratorPlugin/plugin.swift
@@ -11,6 +11,7 @@ import PackagePlugin
 /// Generates the API for the SwiftGodot from the Godot exported Json API
 @main struct SwiftCodeGeneratorPlugin: BuildToolPlugin {
     func createBuildCommands(context: PluginContext, target: Target) throws -> [Command] {
+        var commands: [Command] = []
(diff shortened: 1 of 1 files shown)


tishin:
Disabled `buildCommand` on Swift 5.9 for Windows since 5.9 github action got stuck.

migueldeicaza:
I merged the request- I hope this doesn't break the 5.9 folks for now, but hopefully we will all love to 5.10 soon

tishin:
5.9 still uses original `prebuildCommand` hack without additional `buildCommand`. It shouldn't break.

Issue #317, opened 2023-12-21
Windows: Unable to run macro tests on 5.9.1+ Swift runtime


tishin:
Running `swift test` on Windows using 5.9.1 or 5.9.2 Swift toolchain fails with a linker error:
```
lld-link: error: duplicate symbol: main
>>> defined at C:\\Projects\\SwiftGodot\\.build\\x86_64-unknown-windows-msvc\\debug\\SwiftGodotMacroLibrary.build\\MacroGodot.swift.o
>>> defined at C:\\Projects\\SwiftGodot\\.build\\x86_64-unknown-windows-msvc\\debug\\SwiftGodotPackageTests.build\\runner.swift.o
```
Looks like it picks up `@main` attribute from the compiler plugin - `@main struct godotMacrosPlugin: CompilerPlugin` in addition to the generated `@main` from the package tests runner.

I'll try to find some time to make a simplified example and report this issue to SPM.

tishin:
There is already an issue for this bug: https://github.com/apple/swift-package-manager/issues/7174

PR #425, opened 2024-03-14
Disable macros tests on Windows

tishin:
Macros tests don't work on Windows, as described in #317. This PR disables them for now.


migueldeicaza:
Thank you!

PR #426, opened 2024-03-16
Fix warnings in tests

tishin:
Mainly the actor isolation one: 
`Main actor-isolated class 'GodotTestCase' has different actor isolation from nonisolated superclass 'XCTestCase'; this is an error in Swift 6`
Main actor isolation is not necessary for `GodotTestCase` since those test cases run inside `loadScene` callback, which happens on main thread.


migueldeicaza:
Thank you for this lovely patch

PR #427, opened 2024-03-16
Clean up unnecessary Foundation imports

tishin:
This leaves only 2 imports in `SwiftGodot` where `Foundation` is actually used:
- `PackedByteArray.asData()` returning `Data`
- `Numeric.snapped` native implementation using `floor()`

Both can be refactored without using `Foundation` dependency, which is too heavy for non-Darwin platforms compared to the convenience it brings. I think it should be up to users to decide whether they want to include `Foundation` into their game or not.

### DIFF
--- a/Sources/SwiftGodot/Core/InspectableProperty.swift
+++ b/Sources/SwiftGodot/Core/InspectableProperty.swift
@@ -5,8 +5,6 @@
 //  Created by Marquis Kurt on 10/15/23.
 //
 
-import Foundation
-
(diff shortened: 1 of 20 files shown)


migueldeicaza:
Thank you, I didn’t realize Foundation was the source of this pain.

I don’t mind removing the asData() one for the sake of reducing the dependency.   Would love to find a way of making people’s lives easier still for that common idiom.

migueldeicaza:
I just dropped the `asData` with the suggestion from above.

Issue #429, opened 2024-03-19
[Windows] GodotSwift and SwiftSyntax libraries are being rebuilt every time


daniele-niero:
I have encountered a problem where both GodotSwift and SwiftSyntax libraries are being rebuilt every time I compile my project. 
This makes it extremely hard to develop and test a game on Windows.

The project tested is the SimpleRunnerDriver from the Documentation.

I can understand a slow build the first time, but changing a bit the SimpleRunnerDriver.swift shouldn't trigger the recompilation of SwiftGodot (and by reflection SwiftSyntax)

Swift version: 5.9.2
Operating System: Windows 10
Godot version: 4.2

Is this  a known issue on Windows?

migueldeicaza:
I didn’t know where to start, since I don’t even have a windows machine with me currently, but @tishin found and fixed the issue.

thank you for the report

daniele-niero:
Thank to you guys for the prompt fix.

With Swift 5.9.2 I still had compilation problems and linking that took ages.
I also tried on a different machine and with a newly created project. The problem was still there.

But I just update Swift to 5.10 and now everything seems to work just fine (after the first compilation)

I thought it was worth to mention this.

Thanks again

PR #434, opened 2024-03-20
Fix generator consistency on Windows

tishin:
Addresses #429. Currently the order in which generated sources are put together into a single file on Windows is inconsistent. I'd guess that's the reason why SwiftGodot is being recompiled almost entirely on each build.


migueldeicaza:
This is wonderful!   Thank you for tracking this down!

PR #436, opened 2024-03-20
Rewrite snapped without using Foundation

tishin:
Follow-up to #427. Deals with the last Foundation import.


migueldeicaza:
Woohooo!  We are foundation free!

PR #439, opened 2024-03-21
Proposal: Updated BindNode with safety and convenience lookup

Nemesisprime:
Just a few improvements to make `@BindNode` a little bit better to work with. 

### Improved Lookup

Given a Scene Node named "MovementComponent": 
```swift
    @BindNode var MovementComponent: Node? // existing
    @BindNode var movementComponent: Node? // inferred
    @BindNode("MovementComponent") var customLabel: Node? // custom
```
All become valid ways to look up via BindNode. 

### Safe Lookup

Previously, `@BindNode` didn't include the ability to look-up via optional, but this now forces them to be defined as optionals. This prevents the chance that a misnamed or missing Node is going to cause a crash instead puts it on the programmer to handle the missing node.

This has a backwards incompatible change to more-or-less force safety on BindNode which is primarily the reason why I marked it a proposal in the title here. Feel feel to object but thought I'd take the chance given the early-ish state of the project and how annoying that was to stumble on 🙂

### DIFF
--- a/Sources/SwiftGodot/Extensions/NodeExtensions.swift
+++ b/Sources/SwiftGodot/Extensions/NodeExtensions.swift
@@ -19,25 +19,53 @@
 ///
 ///     getNode(path: NodeName("GameOverLabel")) as! Label
 
+/// Use the BindNode property wrapper in any subclass of Node to retrieve the node from the
+/// current container that matches the name of the property.
+///
+/// For example:
+///
+///     class MyElements: CanvasLayer {
+///         @BindNode var GameOverLabel: Label
+///         @BindNode var gameOverLabel: Label
+///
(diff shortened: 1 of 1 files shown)


ridigilis:
Suggestion: Could the argument label be omitted in the custom case? 

`@BindNode("MovementComponent") var customLabel: Node?`

Normally I favor explicitness, but I think this reads nicely.

Nemesisprime:
Oooh I agree, that looks really nice. Applied the change to the init and docs!

Nemesisprime:
Was speaking in Slack and discovered `@SceneTree` (https://github.com/migueldeicaza/SwiftGodot/pull/297) which more-or-less replaces bindNode with the optional safety behavior (and is tested!). Magic mapping TBD.

Issue #450, opened 2024-04-08
Conflict with custom Variant type when using @Callable


Stiivi:
When a custom Variant type exists, it conflicts with the Godot `Variant` when declaring a `@Callable` function.

I have a code where an imported package contains its own `Variant` type. Normally one would just prefix it with the package name (`MyPackage.Variant`) to remove ambiguity.

The following code snippet will offend the compiler:

```swift
import SwiftGodot

enum Variant {
    case int(Int)
    case string(String)
}

@Godot
class VariantConflictNode: Node {
    @Callable
    public func doSomething() {
        // do nothing
    }
}
```

The errors: 

```
/var/folders/_c/_nk3n3wn3sz_h2vrr17q95200000gn/T/swift-generated-sources/@__swiftmacro_15VariantConflict0aB4Node5GodotfMm_.swift:10:120 Cannot use instance member '_mproxy_doSomething' within property initializer; property initializers run before 'self' is available
```

```
/var/folders/_c/_nk3n3wn3sz_h2vrr17q95200000gn/T/swift-generated-sources/@__swiftmacro_15VariantConflict0aB4Node5GodotfMm_.swift:10:140 Cannot convert value of type '([Variant]) -> Variant?' to expected argument type '(VariantConflictNode) -> ([Variant]) -> Variant?'
```

migueldeicaza:
Thank you for your pull request!

PR #451, opened 2024-04-14
Prevent custom Variant type conflict (#450)

Stiivi:
Prefix `Variant` in GodotCallable macro with `SwiftGodot` to prevent potential conflict with custom `Variant` type.


migueldeicaza:

PR #456, opened 2024-04-25
Expose packed_byte_array_operator_index_const.

kevinw:
Zero-copy access to byte arrays across the gdextension interface is possible with `packed_byte_array_operator_index_const`, which compliments the version we already use in `withUnsafeMutableAccessToData` -- `packed_byte_array_operator_index`

```
// the new GDExtensionInterface method in this PR
let packed_byte_array_operator_index_const: GDExtensionInterfacePackedByteArrayOperatorIndexConst
```

It's useful with a Foundation-using helper method like

```
extension PackedByteArray {
    /// Returns a Data pointing directly at the data in this PackedByteArray
    public func asDataNoCopy() -> Data? {
        withUnsafeConstAccessToData { ptr, count in 
            Data(bytesNoCopy: .init(mutating: ptr), count: count, deallocator: .none)
        }
    }
}
```

### DIFF
--- a/Sources/SwiftGodot/Core/Packed.swift
+++ b/Sources/SwiftGodot/Core/Packed.swift
@@ -81,6 +81,11 @@ extension PackedByteArray {
         return nil
     }
     
+    /// Read-only access the underlying data for this packed byte array
+    ///
+    /// - Parameter method: a callback that is invoked with a non-mutable pointer to the underlying data, and
+    /// the number of bytes in that block of data. The callback is allowed to return nil if it fails to
+    /// do anything with the data.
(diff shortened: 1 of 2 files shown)


kevinw:
> Did the other version make a copy of the data before giving us a pointer?

Yes, if your array's reference count is > 1 -- 

`packed_byte_array_operator_index` ➡️ `&self->ptrw()[p_index]` ➡️ `Vector::_cowdata.ptrw()` ➡️ `CowData::_copy_on_write`

migueldeicaza:
Thank you, I have updated the documentation

PR #457, opened 2024-05-01
Add `return` to `escapeSwift`

jbromberg:
I'm using a godot module that has an enum with `return`. It failed to build since `return` wasn't being escaped. This fixes that.


migueldeicaza:
Nice!

PR #465, opened 2024-05-10
Extend `initSwiftExtension` macro to expose all initialization levels (#464)

### LINKED ISSUE #464: Extending #initSwiftExtension macro to support type registration at editor level
### Motivation
Currently, the `#initSwiftExtension` macro only registers types at a `.scene` initialization level. If you need to expose types that run on editor time (e.g. plugin code), you have to resort to the default boilerplate initialization code, and manually register your editor classes:
```swift
func setupScene (level: GDExtension.InitializationLevel) {
    switch level {
    case .editor:
        register(type: ToolThatWillRunOnEditor.self)
    case .scene:
        // gameplay classes
    default:
        break
    }
}
```

### Solution
Extending the `#initSwiftExtension` to separately register classes on `.editor` init level.
```swift
#initSwiftExtension(
    cdecl: "swift_entry_point",
    editorTypes: [/* editor level classes */],
    sceneTypes: [/* scene level classes */]
)
```

### Questions
1. Considering that this change is dealing with exposing types at different `GDExtension.InitializationLevel` times, is there any interest in also allowing type registration at `.core` and `.servers` levels?
2. Or any reason not to support it?

sekimondre:
closes #464 

Allows registration of types to Godot at all `GDExtension.InitializationLevel` cases.

Originally, this was supposed to support exposing types only at `.editor` initialization level, but the solution was generalized to include all cases. Maybe some esoteric use cases might require exposing types at `.core` and `.servers` levels.

I was able to work out a way of keeping the original declaration intact, so no breaking changes were introduced to the macro API.

Also, I still have to add api documentation and make some little adjustments to the existing one. If the code's good to go already, I'll do it.

### DIFF
--- a/Sources/SwiftGodot/MacroDefs.swift
+++ b/Sources/SwiftGodot/MacroDefs.swift
@@ -119,6 +119,17 @@ public macro initSwiftExtension(cdecl: String,
                                 types: [Wrapped.Type] = []) = #externalMacro(module: "SwiftGodotMacroLibrary",
                                                                         type: "InitSwiftExtensionMacro")
 
+@freestanding(declaration, names: named(enterExtension))
+public macro initSwiftExtension(
+    cdecl: String,
+    coreTypes: [Wrapped.Type] = [],
+    editorTypes: [Wrapped.Type] = [],
+    sceneTypes: [Wrapped.Type] = [],
+    serverTypes: [Wrapped.Type] = []
+) = #externalMacro(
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Brilliant!   Thanks for sending this along!

migueldeicaza:
I added some documentation, let me know if that works, and happy to take additional changes.

PR #468, opened 2024-05-13
Packed: _const interface for subscript getters

kevinw:
Building on #456 - the `subscript` `get` functions for `PackedXYZArray`s do not need writeable pointers from Godot.

This changeset, by using the `_const` versions of the packed array interface functions, can save memory allocations b/c the copy-on-write logic doesn't need to happen for read-only access.

(Some more performance work from my GodotVision profiling—hitting 90FPS is important there!)

### DIFF
--- a/Sources/SwiftGodot/Core/Packed.swift
+++ b/Sources/SwiftGodot/Core/Packed.swift
@@ -17,7 +17,7 @@ extension PackedStringArray {
     /// Accesses a specific element in the ``PackedStringArray``
     public subscript (index: Int) -> String {
         get {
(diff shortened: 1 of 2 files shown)


kevinw:
Also: there's a piece I'm missing here—it would be nice if you had a `PackedVector3Array` with 1000 elements, and calling `map` on it called through the `gdinterface` once, instead of 1000 times. But it's unclear to me where via Swift protocols that happens—I'm assuming it calls `subscript get`, but I'm not sure yet...

migueldeicaza:
This is a lovely optimization, thank you so much for this contribution!

PR #469, opened 2024-05-14
Update WorkingInVsCode.md to include requirement for min platform target

dolapo:
The working in vscode instructions are missing this step to also set the minimum platform target to macOS 13.


migueldeicaza:
Thank you for this fix!

PR #474, opened 2024-05-20
Use Int instead of Int64 across the board, undoes a previous change t…

migueldeicaza:
…o the way we store enumerations, and instead makes it so that our public API that took Int64 values takes Int values


migueldeicaza:
Some caveats, there are some genuine APIs that should be Int64 which this patch undoes, among those APIs to access PackedInt64, so this is not ready.

migueldeicaza:
Ok my main urge to do this is not very important anymore, as now we can export Enums of either type.

PR #479, opened 2024-05-21
Result of getNode is now cached

lorenalexm:
Previously the node would be fetched from the tree each time it was referenced. Meaning this fetch would take place every frame if the property with the `@BindNode` (class property, used within function) was used in `_physicsProcess` or `_process`. The fetching now takes place the first time, and the cached result every time thereafter.


migueldeicaza:

Issue #483, opened 2024-05-24
GArray.append() requires an argument label


rktprof:
**Describe the bug**
When appending to a GArray using GArray.append() you need to supply an argument label like so
`myArray.append(value: myValue)` whereas a regular swift array does not require it

**To Reproduce**
1. Create a GArray and append to it using `.append(myValue)`
2. Compile
3. It will fail with `missing argument label 'value:' in call`

**Expected behavior**
the `GArray.append()` function should function like the regular swift array `.append()` which omits the argument label requirement

PR #485, opened 2024-05-25
Generator Exceptions

joeynelson42:
In preparation to tackle #483 I formalized the generator exception dictionaries into enumerations.

Manually tested to ensure they match the output of the dictionaries.


migueldeicaza:
Thanks for the patch!   I am not sure that I want to introduce the additional architecture just yet, I will cherry pick the change manually.

PR #488, opened 2024-05-30
Explicitly omit argument label for GArray.append

### LINKED ISSUE #483: GArray.append() requires an argument label

joeynelson42:
Third time is the charm!

Fixes #483 by adding `omittedFirstArgLabelList` and checking for any explicitly omitted labels in `BuiltinGen` and `MethodGen`. 

Also deprecated `GArray.append(value: Variant)` and updated internal usages.

### DIFF
--- a/Generator/Generator/BuiltinGen.swift
+++ b/Generator/Generator/BuiltinGen.swift
@@ -388,10 +388,8 @@ func generateBuiltinMethods (_ p: Printer,
         
         for arg in m.arguments ?? [] {
             var eliminate: String = ""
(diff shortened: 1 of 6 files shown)


migueldeicaza:
Glorious!   Thank you!

PR #492, opened 2024-06-13
doc: replace vscode debug method from `attach` to `launch`

MrZak-dev:
a `launch` debug task can be created in vscode, which runs the godot game directly wiht `lldb` debugger.


JosephHill:
Thanks for figuring this out, and for contributing this!  I think this is going to help people a lot, especially debugging on Windows.

That said, I don’t think we want to replace the documentation for attaching to a running game, as I think it is still a useful/common workflow for people to launch the game from Godot.  

 Do you think you could modify your patch so that it maintains the attach documentation, possibly as an additional (or optional) task to include after the launch task?

MrZak-dev:
>  Do you think you could modify your patch so that it maintains the attach documentation, possibly as an additional (or optional) task to include after the launch task?

of course. done!!, and attach is indeed a very helpful method to debug individual scenes .

PR #493, opened 2024-06-14
Fix resolution of dynamic types when only proper supertypes are exposed through GDExtension

konczg:
Originally, using ```Object.method_get_class``` meant that we tried using an object's dynamic type even if it was not registered with GDExtension, which resulted in falling back to the default case of using the static type from the method signature. This PR fixes this by introducing a call to ```gi.object_get_class_name```, which will return the name of the most specialized subtype that is still registered with GDExtension. If this fails for some reason, we fall back to the previous method.


migueldeicaza:
Thank you for your contributioN!

PR #516, opened 2024-08-13
use let instead of var for _initializeClass

mayoff:
In Swift 6, `_initializeClass` as a `static var` produces an error:

> Static property '_initializeClass' is not concurrency-safe because it is non-isolated global shared mutable state

There's no reason to for it to be a `var`, so make it a `let` instead.


migueldeicaza:
Thank you very much!

PR #517, opened 2024-08-13
work around Godot crash when __CFBundleIdentifier isn't set

mayoff:
The method `-[GodotApplicationDelegate applicationDidFinishLaunching:]` (which is part of Godot, not part of SwiftGodot) uses the environment variable `__CFBundleIdentifier` in a way that crashes if the variable is not set.

Terminal sets `__CFBundleIdentifier`, but Xcode does not.

This patch works around that crash.


migueldeicaza:
Thank you for this contribution!

PR #518, opened 2024-08-16
update dependencies to point at swiftlang repos

mayoff:
Apple has been migrating its Swift repos to the swiftlang organization on Github. This patch updates the URLs in Package.swift to point at the new organization.


migueldeicaza:
Oh thank you!

PR #519, opened 2024-08-16
Swift macro testing

mayoff:
This pull request converts all macro expansion tests to use the MacroTesting library from the swift-macro-testing package.

I want to do some cleanup and enhancement of SwiftGodot's macros. In particular I want to make `@Export` automatically infer hints for subclasses of `Node` and `Resource` (both built-in and user-defined) and provide better, type-safe syntax for other hints (like `@Export(.range(0 ... 10, step: 2, .orGreater))` instead of `@Export(.range, "0,10,2,or_greater")`).

To facilitate that work, I propose this patch to migrate the macro tests to a much friendlier testing library. The swift-macro-testing library has several advantages over SwiftSyntaxMacrosTestSupport:

-   It is much easier to create new test cases. The library can update the test case with in the actual macro output the first time a test is run.

-   It is much easier to update test cases when we intentionally change the code generated by a macro. The library can update the test case by replacing the old macro output with the new macro output if you set `record: failed` in the `assertMacro` invocation.

-   It is much easier to verify that diagnostics and fix-its emitted by a macro are attached to the correct part of the input. Check out `testMacroNotOptionalDiagnostic` in `SceneTreeMacroTests.swift` for a great example!

For a full explanation of the design of swift-macro-testing, check out these videos:

https://www.pointfree.co/collections/macros/testing-and-debugging-macros

The video pages include full hand-written transcripts in case you'd rather read than watch.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -114,7 +114,7 @@ targets.append(
             dependencies: [
                 "SwiftGodotMacroLibrary",
                 "SwiftGodot",
-                .product(name: "SwiftSyntaxMacrosTestSupport", package: "swift-syntax")
+                .product(name: "MacroTesting", package: "swift-macro-testing"),
             ]))
 #endif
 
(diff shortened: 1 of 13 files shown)


mayoff:
To be clear, this patch does not add or remove any test cases, or change the inputs or outputs of any test cases other than fixing indentation.

migueldeicaza:
Thanks for this patch, this one will take me more time to review, but I am also a little hesitant on bringing another external dependency.

mayoff:
Perhaps it would be better to ignore this PR until I have some other improvements to offer that build on this change.

PR #520, opened 2024-08-16
eliminate warnings

mayoff:
This patch eliminates all the compiler warninsg I get under Xcode 16.0 beta 4:

PickerNameProviderMacro.swift:58:13
Initialization of immutable value 'types' was never used; consider replacing with assignment to '_' or removing it

RefCounted.swift:24:9
Result of call to 'initRef()' is unused

Array.swift:301:5
@discardableResult declared on a function returning Void is unnecessary

Packed.swift:13:13
'append(value:)' is deprecated: This method signature has been deprecated in favor of append(_:)

Wrapped.swift:143:28
Value 'node' was defined but never used; consider replacing with boolean test

Wrapped.swift:511:17
Result of call to 'retain()' is unused

### DIFF
--- a/Generator/Generator/ClassGen.swift
+++ b/Generator/Generator/ClassGen.swift
@@ -206,7 +206,7 @@ func generateVirtualProxy (_ p: Printer,
 // the mapped name (it is "Array", not "GArray"):
 let discardableResultList: [String: Set<String>] = [
     "Object": ["emit_signal"],
-    "Array": ["append", "resize"],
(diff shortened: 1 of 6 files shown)


mayoff:
Forgot to build the test cases before opening the PR. The force push 9c54b78 eliminates the four warnings from the test cases.

migueldeicaza:
Oh, delightful!  

I'll review tonight!

PR #523, opened 2024-08-18
remove GodotObject protocol

mayoff:
It doesn't seem like the GodotObject protocol has any reason to exist. This patch removes it.

-   The protocol has no requirements.

-   There is exactly one directly conforming type: Object. All descendants of Object inherit the conformance, but any constraint requiring GodotObject conformance could be rewritten to instead require Object inheritance.

-   There are three functions with GodotObject constraints:

    -   `VariantStorable.makeOrUnwrap`
    -   `Variant.asObject`
    -   `lookupObject`

    All test cases pass if I change these constraints to require Object inheritance instead of GodotObject conformance.


mayoff:
apparently all test cases do NOT pass ha ha

Issue #513, opened 2024-08-10
Two sets of leaks


migueldeicaza:
I recently revamped the memory management for SwiftGodot to deal with the RefCounted scenario that needed to have its reference initialized, and it also solved a leak that was caused in code like this:

```
func leak () {
    var img = Image()
}
```

While the new code is a vast design improvement over the previous setup, it regressed a few scenarios that could have been either band-aids or where the ownership protocol contract has been broken, and I need to review those.

# First Leak

```
let node = Node3D()

// Leaks
for _ in 0...10000 {
    for _ in node.getPropertyList() {
        
    }
}

// Doesn't leak
for _ in 0...10000 {
    node.getPropertyList()
}
```

This one is because the indexer is creating a Variant by making a copy of it, but it looks like the indexer call itself already made a copy.  The change was done in the past to solve this issue: https://github.com/migueldeicaza/SwiftGodot/issues/390

# Second Leak

The case is:

```
let bytes = PackedByteArray([137, 80, 78, 71, 13, 10, 26, 10, 0, 0, 0, 13, 73, 72, 68, 82, 0, 0, 0, 1, 0, 0, 0, 1, 1, 3, 0, 0, 0, 37, 219, 86, 202, 0, 0, 0, 3, 80, 76, 84, 69, 0, 0, 0, 167, 122, 61, 218, 0, 0, 0, 1, 116, 82, 78, 83, 0, 64, 230, 216, 102, 0, 0, 0, 10, 73, 68, 65, 84, 8, 215, 99, 96, 0, 0, 0, 2, 0, 1, 226, 33, 188, 51, 0, 0, 0, 0, 73, 69, 78, 68, 174, 66, 96, 130])

for _ in 0...500000 {
    let resource = Variant(SwiftGodot.Image()).asObject(SwiftGodot.Image.self)!
    _ = resource.loadPngFromBuffer(bytes)
    // Doesn't leak with line below uncommented
    // resource.unreference()
}
```

This is caused by the additional call to `reference` in `asObject`:

```
    public func asObject<T:GodotObject> (_ type: T.Type = T.self) -> T? {
        guard gtype == .object else {
            return nil
        }
        var value: UnsafeRawPointer = UnsafeRawPointer(bitPattern: 1)!
        toType(.object, dest: &value)
        if value == UnsafeRawPointer(bitPattern: 0) {
            return nil
        }
        let ret: T? = lookupObject(nativeHandle: value)
[...]

PR #524, opened 2024-08-19
Add tests cases for, and fix, #513

mayoff:
This patch addresses both leaks discussed in #513.

### First leak

The changes in https://github.com/migueldeicaza/SwiftGodot/commit/d1d6ca47f0fc5bb16948017e1a6b2c6c8838b3b5 made several `Geometry2DTests` fail.

The actual unbalanced retain causing the leak in `VariantRepresentable.swift`:

```swift
extension ContentVariantRepresentable {
    public init? (_ variant: Variant) {
        guard Self.godotType == variant.gtype else { return nil }

        var content = Self.zero
        withUnsafeMutablePointer(to: &content) { ptr in
            variant.toType(Self.godotType, dest: ptr) // <-- THIS IS THE UNBALANCED RETAIN
        }

        self.init(content: content)
    }
}
```

The `variant.toType` function calls into Godot to copy out the built-in type stored in `variant`. If that built-in type has an internal reference count, Godot increments it. Then, the call to `self.init(content: content)` increments the count again. The retain performed by `variant.toType` is never balanced.

This patch fixes the bug by generating a new `init(alreadyOwnedContent:)` initializer for each Godot builtin class type. This needs to be on the builtin wrappers (like `GDictionary`, `GArray`, etc.), rather than on `Variant` which is where https://github.com/migueldeicaza/SwiftGodot/commit/d1d6ca47f0fc5bb16948017e1a6b2c6c8838b3b5 put it.

This patch also adds a test case to check for the leak by looking at Godot's `MEMORY_STATIC` performance counter, which is only enabled if Godot was built with `DEBUG_ENABLED`.

### Second leak

This patch adds a test for the second leak described in https://github.com/migueldeicaza/SwiftGodot/issues/513, and fixes the leak.

Without this patch, the leak happens because `Variant.asObject` has an unneeded call to `rc.reference()` which increments the `RefCounted` object's reference count. As far as I can tell, nothing balances this increment.

`Variant.asObject` calls `lookupObject(nativeHandle:)`, which returns a Swift wrapper for an object whose reference count has already been incremented (if the object is `RefCounted`). The Swift wrapper balances that increment with a decrement in its `deinit`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ build-docs.log
 *rej
 *orig
 \#*
+/libgodot.xcframework/
--- a/Generator/Generator/BuiltinGen.swift
+++ b/Generator/Generator/BuiltinGen.swift
@@ -643,6 +643,10 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
                         p ("\(typeName).constructor1 (&self.content, &args)")
                     }
                 }
+                p ("// Used to construct objects when the underlying built-in's ref count has already been incremented for me")
+                p ("public required init(alreadyOwnedContent content: ContentType)") {
+                    p ("self.content = content")
+                }
             }
            
             func memberDoc (_ name: String) {
--- a/Package.swift
+++ b/Package.swift
@@ -120,15 +120,32 @@ targets.append(
 
 // libgodot is only available for macOS
 #if os(macOS)
+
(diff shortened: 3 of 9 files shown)


migueldeicaza:
You are absolutely right about these observations, I ran into these while cooking the tests, and noticed the 2D tests fail and spent some quality time trying to figure it out, when I saw that I decided to check on the bug again and saw your analysis.

Love your patch, and the additional insight on the differences on those built-in types.

PR #529, opened 2024-08-23
die if a class name has already been registered

mayoff:
I noticed this comment:

```
    // TODO: what happens if the user subclasses but the name conflicts with the Godot type?
    // say "class Sprite2D: Godot.Sprite2D"
```

The only way SwiftGodot can find out the type for some pointer passed from Godot is by checking the class name, we really can't allow class name conflicts: we could end up casting a pointer to the wrong type, leading to difficult-to-diagnose crashes. I think it's better to crash immediately if the extension tries to register a class name that's already been registered.


migueldeicaza:
Good call.

PR #530, opened 2024-08-25
Fix deprecation warning for @Callable

lupuchard:
Fixes the warning I get with the @Callable macro:
` 'append(value:)' is deprecated: This method signature has been deprecated in favor of append(_:)`


migueldeicaza:

PR #532, opened 2024-08-30
Fix macro tests expecting deprecated `append` API

elijah-semyonov:
After `value` label was omitted in https://github.com/migueldeicaza/SwiftGodot/pull/488, the tests macro tests were still expecting the deprecated API. This PR fixes those tests failing.


migueldeicaza:
Thank you very much!

PR #533, opened 2024-09-02
Allocation-less marshaling for built-in and class types methods, and `Callable` changes

elijah-semyonov:
This PR contains improvements to avoid allocations, based on the observation most of those allocations are transient and data can be kept on the stack

1. For marshalled methods of built-in types and classes - keep everything on stack, avoid using `[UnsafeRawPtr?]`. Old code is still available using `LEGACY_MARSHALLING` define, which is currently commented-out. Performance difference can be tested using `testBuiltinsTypesMethodsPerformance` and  and `testClassesMethodsPerformance` in `MarshalTests` with this flag on and off.
2. For `Callable` API - change `[Variant]` to new `Arguments` type, that either contains a Swift array if a call was issued from the Swift side, or wraps a Godot managed `pargs` and lazily reconstructs `Variant`s inside when requested to avoid eager array initialisation. Guarantee of it not becoming dangling is based on the fact that `Arguments` is `~Copyable` and only visible on the user side as `borrowing`.

### DIFF
--- a/Generator/Generator/Arguments.swift
+++ b/Generator/Generator/Arguments.swift
@@ -300,8 +300,8 @@ func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasRetu
             getArgRef(arg: arg)
         }.joined(separator: ", ")
         
-        body += "\(prefix)\(retFromWith)withUnsafePointerToUnsafePointers(\(argsString)) { nPtrs in nPtrs.withMemoryRebound(to: UnsafeRawPointer?.self, capacity: \(args.count)) { args in"
-        withUnsafeCallNestLevel += 2
+        body += "\(prefix)\(retFromWith)withUnsafeArgumentsPointer(\(argsString)) { args in"
+        withUnsafeCallNestLevel += 1
     }
     
     return (body, withUnsafeCallNestLevel)
--- a/Generator/Generator/Arguments.swift
+++ b/Generator/Generator/Arguments.swift
@@ -232,12 +232,33 @@ func generateCopies (_ args: [JGodotArgument]) -> String {
     return body
 }
 
-#if LEGACY_MARSHALING || !canImport(Darwin)
+/// Wrap arguments before marshaling them to Godot
+///
+/// Returns a tuple of
+/// - `body` - blob with arguments preparation
+/// - `identation` - for caller to add missing braces
+/// - `argsRef` - literal expression to be passed into GDExtensionInterface. Can be `nil` if method takes no args, `args` if arguments are wrapped into stack-resident struct, or `&args` if legacy approach is used and `[UnsafeRawPointer?]` is passed.
+func generateArgPrepare(isVararg: Bool, _ args: [JGodotArgument], methodHasReturn: Bool) -> (body: String, indentation: Int, argsRef: String) {
+    // A runtime hook for opt-out from using new marshaling in case some specific method breaks
+    var useLegacyMarshalling = false
+        
+    #if LEGACY_MARSHALING || !canImport(Darwin)
+    useLegacyMarshalling = true
+    #endif
+    
+    if useLegacyMarshalling {
+        return generateArgPrepareLegacy(isVararg: isVararg, args, methodHasReturn: methodHasReturn)
+    } else {
+        return generateArgPrepareNew(isVararg: isVararg, args, methodHasReturn: methodHasReturn)
+    }
+}
 
-func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasReturn: Bool) -> (String, Int) {
+func generateArgPrepareLegacy(isVararg: Bool, _ args: [JGodotArgument], methodHasReturn: Bool) -> (body: String, indentation: Int, argsRef: String) {
     var body = ""
     var withUnsafeCallNestLevel = 0
     let retFromWith = methodHasReturn ? "return " : ""
+    
+    let argsRef: String
     
     if isVararg || args.count > 0 {
         body += generateCopies (args)
@@ -259,16 +280,21 @@ func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasRetu
             body += "    args.append (cptr.baseAddress! + idx)\n"
             body += "}\n"
         }
+        
+        argsRef = "&args"
+    } else {
+        argsRef = "nil"
     }
-    return (body, withUnsafeCallNestLevel)
+    
+    return (body, withUnsafeCallNestLevel, argsRef)
 }
 
-#else
-
-func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasReturn: Bool) -> (String, Int) {
+func generateArgPrepareNew(isVararg: Bool, _ args: [JGodotArgument], methodHasReturn: Bool) -> (body: String, indentation: Int, argsRef: String) {
     var body = ""
     var withUnsafeCallNestLevel = 0
     let retFromWith = methodHasReturn ? "return " : ""
+    
+    let argsRef: String
     
     // TODO: this case should get the same treatment as a second branch.
     if isVararg {
@@ -286,7 +312,9 @@ func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasRetu
         body += "for idx in 0..<arguments.count {\n"
         body += "    cptr [idx] = arguments [idx].content\n"
         body += "    args.append (cptr.baseAddress! + idx)\n"
-        body += "}\n"        
+        body += "}\n"
+        
+        argsRef = "&args"
     } else if args.count > 0 {
         body += generateCopies (args)
         
@@ -298,9 +326,11 @@ func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasRetu
         
         body += "\(prefix)\(retFromWith)withUnsafeArgumentsPointer(\(argsString)) { args in"
         withUnsafeCallNestLevel += 1
+        
+        argsRef = "args"
+    } else {
+        argsRef = "nil"
     }
     
-    return (body, withUnsafeCallNestLevel)
+    return (body, withUnsafeCallNestLevel, argsRef)
 }
-
-#endif
--- a/Generator/Generator/Arguments.swift
+++ b/Generator/Generator/Arguments.swift
@@ -7,6 +7,8 @@
 
 import Foundation
 import ExtensionApi
+import SwiftSyntax
+import SwiftSyntaxBuilder
 
 func godotArgumentToSwift (_ name: String) -> String {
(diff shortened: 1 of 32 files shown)


migueldeicaza:
I completed a first pass of the patch, in general, it feels directionally fine, but I need to digest the changes and truly understand what the changes do, luckily the history of the commits might shed some light into that.

It is vast, so I am a little nervous about breaking things, so I will need to test locally, and test against Godot for iPad that exercises SwiftGodot extensively in the next few days.

elijah-semyonov:
Rewrote commit history and cleaned up to make it easier to follow!

Generator/Generator/Arguments.swift
```diff
@@ -258,3 +262,49 @@ func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasRetu
...
+
+#else
+
+func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasReturn: Bool) -> (String, Int) {
+    var body = ""
+    var withUnsafeCallNestLevel = 0
+    let retFromWith = methodHasReturn ? "return " : ""
+    
+    // TODO: this case should get the same treatment as a second branch.
+    if isVararg {
```
migueldeicaza, line 274: Could you explain what you had in mind for this?

The original code has this condition:

```
isVararg || args.count > 0
```

For this block, and here it seems to be handled differently, so I am curious for the rationale here.

There is additionally some check duplication here, which was necessary on the original code, but now looks redundant (the nested `if isVararg`):

Generator/Generator/Arguments.swift
```diff
@@ -258,3 +262,49 @@ func generateArgPrepare (isVararg: Bool, _ args: [JGodotArgument], methodHasRetu
...
+        body += generateCopies (args)
+        body += "var args: [UnsafeRawPointer?] = []\n"
+        if isVararg {
+            body += "let cptr = UnsafeMutableBufferPointer<Variant.ContentType>.allocate(capacity: arguments.count)\n"
```
migueldeicaza, line 278: This is the test duplication, and there is another one below from the original code.

Generator/Generator/BuiltinGen.swift
```diff
@@ -259,8 +269,9 @@ func generateMethodCall (_ p: Printer,
             p ("\(typeName).\(methodToCall) (nil, \(ptrArgs), \(ptrResult)\(numberOfArgs))")
     } else {
         if isStructMap [typeName] ?? false {
-            p ("withUnsafePointer (to: self) { ptr in ")
-            p ("    \(typeName).\(methodToCall) (UnsafeMutableRawPointer (mutating: ptr), \(ptrArgs), \(ptrResult)\(numberOfArgs))")
+            p ("var mutSelfCopy = self")
+            p ("withUnsafeMutablePointer (to: &mutSelfCopy) { ptr in ")
+            p ("    \(typeName).\(methodToCall) (ptr, \(ptrArgs), \(ptrResult)\(numberOfArgs))")
             p ("}")
```
migueldeicaza, line 275: Question, what was the rationale for this change?  

We have two scenarios, one, Godot will never mutate on these calls, so in that case, this seems unnecessary.   But if there is ever a mutation of the code, then this would discard the mutation.   So I want to understand this, as it seems like this could be a semantic change that could become tricky to diagnose.

Generator/Generator/BuiltinGen.swift
```diff
@@ -601,8 +612,8 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
             if bc.name == "Callable" {
                 p ("/// Creates a Callable instance from a Swift function")
                 p ("/// - Parameter callback: the swift function that receives an array of Variant arguments, and returns an optional Variant")
-                p ("public init (_ callback: @escaping ([Variant])->Variant?)") {
-                    p ("content = CallableWrapper.makeCallable (callback)")
+                p ("public init(_ callback: @escaping (borrowing Arguments) -> Variant?)") {
+                    p ("content = CallableWrapper.callableVariantContent(wrapping: callback)")
                 }
```
migueldeicaza, line 617: This breaks the existing API, is there a reason to not introduce this as an additional constructor/optimization instead of breaking existing code?

Generator/Generator/UnsafePointerHelpers.swift
```diff
@@ -0,0 +1,96 @@
...
+///
+///     return withUnsafePointer(to: &storage) { ptr in
+///         ptr.withMemoryRebound(to: UnsafeRawPointer?.self, capacity: 3) { rawPtr in
+///             body(rawPtr)
+///         }
+///     }
+/// }
+/// ```
+private func generateWithUnsafeArgumentsPointer(argumentsCount count: Int) -> String {
+    let funcDecl = FunctionDeclSyntax(
```
migueldeicaza, line 55: I wonder if we should flag these as: `@inline(__always)`

Sources/SwiftGodot/Core/ClassServices.swift
```diff
@@ -133,13 +133,14 @@ public class ClassInfo<T:Object> {
         if let returnValue {
             retInfo = returnValue.makeNativeStruct()
         }
-        let functionInfo = Unmanaged.passRetained(FunctionInfo (function, retType: returnValue?.propertyType))
+        let userdata = UnsafeMutablePointer<FunctionInfo>.allocate(capacity: 1)
+        userdata.initialize(to: .init(function, retType: returnValue?.propertyType))
```
migueldeicaza, line 138: I am a little nervous that we used to pass a retained value, and now we do not.

That said, it looks like we never released the value in the past.   Maybe it was always persisted.

Sources/SwiftGodot/Core/Wrapped.swift
```diff
@@ -539,57 +539,52 @@ func frameworkTypeBindingFree (_ token: UnsafeMutableRawPointer?, _ instance: Un
 }
 
 /// This function is called by Godot to invoke our callable, and contains our context in `userData`,
-/// pointer to Variants, an argument count, and a way of returing an error
```
migueldeicaza, line 542: This took me a while to digest, but it does look fine.

Generator/Generator/Arguments.swift
```diff
@@ -230,6 +232,8 @@ func generateCopies (_ args: [JGodotArgument]) -> String {
     return body
 }
 
+#if LEGACY_MARSHALING
+
```
migueldeicaza, line 236: I also wonder if we could make this a runtime feature rather than a compile time feature.

The reason being, that way, if we find that for a specific method we need the legacy option, we could turn that on/off.

migueldeicaza (commented):
This looks fabulous, and I have included here a diff with the changes to the generated code:

https://gist.github.com/migueldeicaza/701c657f1a5ece15b177739a5e69000d

Would love your comments to my notes in the patch.

Generator/Generator/BuiltinGen.swift
elijah-semyonov: The rationale behind this change is that just like const-cast (which `.init(mutating:)` is) the a mutation done by mistake by Godot won't be _just_ discarded, it's in fact a UB.
May be it's indeed super-defensive programming.

Generator/Generator/BuiltinGen.swift
elijah-semyonov: The main reason for removing the older one is that 

```swift
let callable = Callable { args in 
    // ...
}
```

becomes ambiguous. Depending on how people accessed the old args (by not passing the lambda from somewhere else and just using indexing operator) it can be potentially not source-breaking

Generator/Generator/UnsafePointerHelpers.swift
elijah-semyonov: Yeah, I can rerun benchmarks and see how much improvement it brings.

Sources/SwiftGodot/Core/ClassServices.swift
elijah-semyonov: Yeah, it's a registration thing and always persisting. Using `Unmanaged` is an overkill here, imo. Even if we bring deregistration logic, it would be enough just to deinitialize and deallocate the `FunctionInfo`

Generator/Generator/Arguments.swift
elijah-semyonov: Hm, perhaps we also could check if there is a way in SPM to propagate some `define` to dependency libraries?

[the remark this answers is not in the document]
elijah-semyonov: Swift 6.0 was released just now so I'll check if the issues was resolved

Generator/Generator/Arguments.swift
elijah-semyonov: Yes, it's indeed a redundant check now. That's an oversight, will fix.

What I had in mind is that I still haven't figured out the difference between `isVararg` being true and false. So when it's true, the original implementation is used.

For a case when it's false and there are arguments, I use the new approach with on-stack storage.

elijah-semyonov: Fixed

Generator/Generator/Arguments.swift
elijah-semyonov: Fixed

Generator/Generator/Arguments.swift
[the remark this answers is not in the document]
elijah-semyonov: It made things worse, added `|| !canImport(Darwin)`

Generator/Generator/UnsafePointerHelpers.swift
elijah-semyonov: There is no measurable change in runtime performance on my machine.

Generator/Generator/Arguments.swift
elijah-semyonov: Ah, I misunderstood you. You probably meant in the scope of CodeGen.

elijah-semyonov: Done in [8e5edac](https://github.com/migueldeicaza/SwiftGodot/pull/533/commits/8e5edacf43bfbb383c48514a1eec1bc7f6c7c88b)

Generator/Generator/BuiltinGen.swift
elijah-semyonov: I returned it back and marked it as deprecated. Do you think it's a viable approach?

migueldeicaza: I think this works, I will try it with my big codebase now and report back, we might need to tune it in the future.

Generator/Generator/Arguments.swift
migueldeicaza: Thank you!  This is brilliant!

migueldeicaza:
Thank you so much for this contribution and the various iterations to get it to this new state.

Cheers!

PR #534, opened 2024-09-08
Update WINDOWS_BUILD.md

jay18001:
Added information about Swift Windows DLLs in the windows build file


migueldeicaza:
Thanks!  Question, are all those libraries necessary?  Or are those just necessary if those features are used?

jay18001:
I think they are all required because they are from the standard library. Linux will probably need the Linux libraries too.

PR #536, opened 2024-09-12
Add additional reserved words to generator

jakobwinkler:
While experimenting with customized extension APIs, I came across these identifiers that were not escaped previously.

Specifically, these 4 (`switch`, `public`, `private`, `for`) are used in <https://github.com/GodotSteam/GodotSteam>.


migueldeicaza:
Thanks for your contribution!

PR #540, opened 2024-09-23
A minor dev UX Generator improvement

elijah-semyonov:
When running generator without any parameter from either terminal or Xcode - it will use repo-local gitignored folder for output by default and repo-local Godot API JSON


migueldeicaza:
Thank you for this change!

Issue #541, opened 2024-09-23
String leak when wrapping in a Variant


migueldeicaza:
It looks like we have a string leak when wrapping in a variant:

![image](https://github.com/user-attachments/assets/70ede0dc-f467-4730-8e94-77a75b9a8d0c)

Doing this:

```
Variant("foo")
```

Sample repro:

```
for _ in 0...1000000 {
    let string = Variant("foo")
}
```

PR #542, opened 2024-09-24
Fix a 541 leak

elijah-semyonov:
Fixes https://github.com/migueldeicaza/SwiftGodot/issues/541

Godot String (aka GString) is described as `"has_destructor": true`. In the case of the bug, it was created, Variant made a copy of it that it manages by itself, but it was never destructed. `GString` would manage its destruction in such a case. `GStringRaw` was removed as redundant.


migueldeicaza:
Thank you for this fix!

PR #545, opened 2024-09-24
Fix a 544 leak

elijah-semyonov:
Fixes https://github.com/migueldeicaza/SwiftGodot/issues/544:

`variant_call` returns a retained copy, we don't need to `copy` it, but assume control over it instead.


migueldeicaza:
Thank you so much!

PR #546, opened 2024-09-25
Some of the custom builtin vector3 operators and members

elijah-semyonov:
Reimplement some of builtin operators of Vector3 in Swift to avoid marshaling


migueldeicaza:
Oh I love that this approach keeps the documentation!

Insta merge!

Issue #543, opened 2024-09-24
Godot String via Variant subscript operator leaks


elijah-semyonov:
Repro:

```
let string = "Hello, World!"
let variant = Variant(string)

for _ in 0 ..< 2000 {
    let retrieved = variant[0]
}
```

Issue #531, opened 2024-08-27
Variant crash


migueldeicaza:
The following example that implements an encoding for Godot crashes, due to attempting to access freed information:

```
//
//  main.swift
//
// Very trivial example of using SwiftGodotKit
//
//  Created by Miguel de Icaza on 4/1/23.
//

//import Foundation
import SwiftGodot
import SwiftGodotKit

struct Foo: Codable {
    var myInt: Int
    var myText: String
    var myIntArray: [Int]
}

struct FooN: Codable {
    var myInt: Int
    var myText: String
    var myFoo: [Foo]
}

protocol GodotEncodingContainer {
    var data: Variant { get }
}

extension Vector2: Codable {
    enum CodingKeys: String, CodingKey {
        case x
        case y
    }

    public init (from decoder: Decoder) throws {
        let values = try decoder.container(keyedBy: CodingKeys.self)
        let x = try values.decode(Float.self, forKey: .x)
        let y = try values.decode(Float.self, forKey: .y)
        self.init (x: x, y: y)
    }

    public func encode(to encoder: any Encoder) throws {
        var container = encoder.container(keyedBy: CodingKeys.self)
        try container.encode(x, forKey: .x)
        try container.encode(x, forKey: .y)
    }
}

/// Notes on encoding:
///   - Int8 is encoded as an Int
public class GodotEncoder: Encoder {
    public var codingPath: [any CodingKey] = []
    public var userInfo: [CodingUserInfoKey : Any] = [:]
    var container: GodotEncodingContainer? = nil

    public init () {

    }

    public func container<Key>(keyedBy type: Key.Type) -> KeyedEncodingContainer<Key> where Key : CodingKey {

        var container = GodotKeyedContainer<Key>(codingPath: codingPath, userInfo: userInfo)
        self.container = container
        return KeyedEncodingContainer(container)
    }

    public var data: Variant {
        return container?.data ?? Variant()
    }

    func encode(key codingKey: [CodingKey], value: Variant) {
        let key = codingKey.map { $0.stringValue }.joined(separator: ".")
        fatalError()
        //dict [key] = value
    }

    public func unkeyedContainer() -> any UnkeyedEncodingContainer {
        let container = GodotUnkeyedContainer(codingPath: codingPath, userInfo: userInfo)
        self.container = container
        return container
    }

    public func singleValueContainer() -> any SingleValueEncodingContainer {
[...]

PR #548, opened 2024-09-26
Revisit indexing operators and fix leaks and crashes

elijah-semyonov:
Fix https://github.com/migueldeicaza/SwiftGodot/issues/543 and https://github.com/migueldeicaza/SwiftGodot/issues/531
Reduce performance checks iterations x10.  
Fix dictionary leaks and a crash when `dictionary[key] = nil`


Sources/SwiftGodot/Variant.swift
```diff
@@ -273,7 +273,8 @@ public class Variant: Hashable, Equatable, CustomDebugStringConvertible {
             if valid == 0 || oob != 0 {
                 return nil
             }
-            return Variant(copying: _result)
+                        
+            return Variant(takingOver: _result)
```
elijah-semyonov, line 277: Should probably add a comment here. Unlike `Array`, that gives a pointer to its internal storage, this API returns a copy.

PR #553, opened 2024-10-01
Even fancier marshaling

elijah-semyonov:
Use nested `withUnsafePointer` so that Windows toolchain doesn't die.
Utilize an ability of `withUnsafePointer` to create its own internal copies.
Use `withUnsafeTemporaryAllocation` for variadics.
Use on-stack strategy when possible (non-variadics and empty variadics).
Unify BuiltInGen and MethodGen generation.
Lots of minor refactorings.
Remove redundant Variant copies in some cases.


migueldeicaza:
Some quick observations:

`final` seems to have changed in the public API:

From:

```
public final func start (mode: AESContext.Mode, key: PackedByteArray, iv: PackedByteArray = PackedByteArray())-> GodotError {
```

to:

```
public func start(mode: AESContext.Mode, key: PackedByteArray, iv: PackedByteArray = PackedByteArray()) -> GodotError {
```

elijah-semyonov:
It was a small nasty oversight!
Already fixed in https://github.com/migueldeicaza/SwiftGodot/pull/553/commits/442045515711a053e31e860f8ba58e7eef61b6e1

PR #560, opened 2024-10-03
Update to 4.3

migueldeicaza:


migueldeicaza:
So this is very stressful, the tests pass locally from Xcode, but they fail with "swift test" on the command line.

PR #565, opened 2024-10-04
Support for new types in godot 4.4.  Added better error messages for …

jordo:
Should be mergeable to `main` and `4.3`.  Will be needed for 4.4 generator eventually.


migueldeicaza:
There were a couple of conflicts with the 4.3 merge, but I merged an equivalent version.

PR #567, opened 2024-10-05
Eradicate optional Variant

elijah-semyonov:
Change API to reflect nullability as an intrinsic part of `Variant` semantics to avoid awkward states like `Variant?` containing `.some(Variant)` containing Godot `Nil`.


migueldeicaza:
You have some enthusiastic comments on Slack during your code review that lead to this, would love to know more about those.

generally, being a breaking change I am a bit concerned about this, but also losing the language support.  An alternative is going the other way, but as I mentioned on slack, that could lead to an impedance mismatch.

so I think it would be good to explore the pros and cons in depth before crossing either bridge

elijah-semyonov:
I'll cook the opposite approach and try to amalgamate `Godot` nil into Swift nil.

PR #578, opened 2024-10-17
Add an entry point generator plugin and documentation on how to use it

elijah-semyonov:
The plugin, when used with a target - will generate an entry point, registering all classes with `@Godot` macro attached.

Screenshot 2024-10-17 at 17 31 20

Screenshot 2024-10-17 at 17 31 49


README.md
```diff
@@ -175,6 +175,30 @@ import SwiftGodot
 #initSwiftExtension(cdecl: "swift_entry_point", types: [SpinningCube.self])
 ```
 
+Also, you can use `EntryPointGeneratorPlugin` that will scan the target source files and generate entry point called `swift_entry_point` with `types` array mentioning all classes with `@Godot` macro attached. All you need is to add `plugins` entry in your `Package.swift` as below:
```
samdeane, line 178: I'd be tempted to rewrite the README to make using the plugin the recommended way of doing things.

You could tell people to only define their own entry point if they are doing something funky. Could possibly even remove the macro at that point.

I'm guessing that people forgetting to add their type to the list is a common source of confusion.

elijah-semyonov: On the other side, it certainly has an overhead of all target files top-level AST walk and other hidden overhead related to plugin environment. We can always change it later.

migueldeicaza:
Lovely!

Issue #537, opened 2024-09-20
Receiving editor warning for GArray - Could not find element type from property hint of a typed array


diiiv9:
I am receiving the continuous warnings in the Godot console when I include a Swift function that returns a `GArray`.

Swift Method (ignore build error) : 
<img width="527" alt="Screenshot 2024-09-20 at 12 30 02 PM" src="https://github.com/user-attachments/assets/aee6d629-c352-4368-8000-987f83058e33">

When calling the above Swift method in Godot , editor gives the _warning_ :
<img width="520" alt="Screenshot 2024-09-20 at 12 31 18 PM" src="https://github.com/user-attachments/assets/e3527763-dec8-4daf-a52b-d4154b97f5d1">

This is just an editor warning , _everything builds and run fine_.

I notice that in the generated documentation , the return type is mentioned as `Array[]` , elsewhere this is usually just `Array`.

<img width="784" alt="Screenshot 2024-09-20 at 12 32 00 PM" src="https://github.com/user-attachments/assets/5cda399f-848a-42db-a4eb-e4c61f838986">

I am using **Godot 4.4 dev2** . I haven't tested the issue on other builds.

elijah-semyonov:
I think I fixed it :)

diiiv9:
@elijah-semyonov I am still receiving this error in 4.4.

elijah-semyonov:
Are you on master?

elijah-semyonov:
Can you paste the Swift counterpart for the one causing the issue?

diiiv9:
<img width="551" alt="Image" src="https://github.com/user-attachments/assets/425e24ff-316e-434e-8271-94c29e7a2f66" />

PR #581, opened 2024-10-19
Don't add `.arrayType` hint when registering `@Callable` methods returning `GArray`

elijah-semyonov:
As per doc, Godot will expect a type name in `hintStr`, in our case it was en empty string that was treated as "TypedArray of type with empty name".

Now the methods are properly exported:
Screenshot 2024-10-19 at 13 49 52
Screenshot 2024-10-19 at 13 49 14

Fixes: https://github.com/migueldeicaza/SwiftGodot/issues/537


migueldeicaza:
Beautiful!  Thank you!

Issue #42, opened 2023-05-01
Cluster generated signals


migueldeicaza:
Right now to get the ball rolling I just create helper signal classes for each signal.

The signal class could be named after the signal emitted, and in cases where we have clusters of identical parameters, we could try to come up with a name that works.  My initial review shows that this would work fine.

Just need time to write the code

samdeane:
I think that the variadic arguments can be constrained to `VariantRepresentable`, so that the emit call can just take normal swift types, and the connect closure can receive normal types.

samdeane:
This should allow us to stop generating any helper classes - we just generate a computed property for each signal which returns a helper.

If this works, then adopting this helper should be a transparent change which doesn't break anything.

The same helper could then be used to implement a cleaner mechanism for declaring signals in Swift (see #584) -- which might be a breaking change.

samdeane:
Unless this seems like a bad idea, I will experiment with the approach above, and see if I can get it to fly. 

Will raise a PR if I get anywhere...

samdeane:
Implementation done - just testing it a bit...

migueldeicaza:
Closing.

PR #584, opened 2024-10-23
Unify Swift and Godot Signal Syntax

samdeane:
See also #586, which describes the problem I was trying to solve.

This PR builds on the changes in #587 - which should be tested and merged first.

Motivation
The #signal macro defines the signal as a static property on the class that contains it.

This means that, given a class with a signal:

@Godot public class SBActor: Object {
    #signal("weapons_changed")
    // ...etc
}
when using the signal in SwiftGodot code, the current spelling to connect to it is:

let actor = SBActor()
actor.connect(signal: SBActor.weaponsChanged, to: self, method: "updateWeapons")
This isn't as intuitive as the GDScript spelling, and it's different from the way we treat signals that are defined in Godot and bridged over to Swift, eg:

myNode.ready.connect {
    print ("The node is ready")
}
Implementation
This PR adds a new #nusignal macro expansion which generates a computed property on the class, using GenericSignal. This is the same approach that is used to implement the signals imported from the Godot API.

We also add an emit() method to GenericSignal, which is of course essential to support emitting signals defined in Swift.

This allows the swift connect call to be written as:

actor.weaponsChanged.connect { print("do something here"); }
It can also be written as:

actor.weaponsChanged.connect(handleWeaponsChanged)

func handleWeaponsChanged() { print("do something here") }
To Do
I added a new macro to ensure that this isn't a breaking change.

Instead of the #xxx style macro, it might be possible to make an attachment-style@Signal, which could be cleaner

If this change is adopted, it's probably worth updating the documentation and deprecating the old macro.


migueldeicaza:
We have an existing idiom to map Godot signal into Swift idioms, have you seen it?

I wonder if we can reuse that, so we do not end up with an additional idiom - check the generated code we have.

samdeane:
> We have an existing idiom to map Godot signal into Swift idioms, have you seen it?

Interesting. I thought I'd read all the relevant code, but I'm not exactly sure what you're referring to.

samdeane:
Ah, are you talking about the `SimpleSignal` stuff in `SignalSupport.swift`?

That looks like it supports the syntax I prefer, but only for zero-argument signals.

samdeane:
I guess you're talking about support for the syntax described [here](https://migueldeicaza.github.io/SwiftGodotDocs/documentation/swiftgodot/signals#Using-Signals)?

It confused me that this syntax seems to work for signals declared outside of SwiftGodot (eg existing signals on Node), but it doesn't work for signals we declare using ourselves using `#signal`.

It would indeed make sense to try to use the same underlying mechanism for both. It feels like it should be achievable, and in a way that is less complex/repetitive than the current `#signal` implementation.

samdeane:
Ignoring the optimisations for now, I wonder if I can lift the code from `generateSignalType` into a place where the `#signal` macro could also call it. 

Then using `#signal("foo", Bar.self)`

 could emit a `SignalFoo` support class and a computed property:

```swift
    public class SignalFoo { /*. generated support for Foo signal here... */ }
    public var foo: SignalFoo { SignalFoo(target: self, signalName: "foo") }
```

and you could do `obj.foo.connect { bar in ... }`, or `obj.foo.emit(Bar())` like you can for the API signals.

samdeane:
A mockup of a generic Signal implementation...

```swift
import Testing

@testable import Scratch

@Test func name() async throws {
  struct Foo {
    var foo: GenericSignal<Int> { GenericSignal<Int>(name: "foo") }
  }

  let f = Foo()
  f.foo.connect { (i: Int) in
    print("got \(i)")
  }

  f.foo.emit(42)
}

struct GenericSignal<each T> {
  let name: String

  func emit(_ t: repeat each T) {
    emitSignal(name, repeat each t)
  }

  func connect(callback: @escaping (_ t: repeat each T) -> Void) {
    var sp = SignalProxy()
    sp.proxy = { args in
      callback(repeat args.pop(as: (each T).self)!)
    }
  }
}

class ArgArray {
  var args: [Any] = []
  func pop<T>(as: T.Type) -> T? {
    let v = args.first as? T
    args.removeFirst()
    return v
  }
}

struct SignalProxy {
  var proxy: ((ArgArray) -> Void)? = nil
}

func emitSignal<each T>(_ name: String, _ args: repeat each T) {
  let a = (repeat each args)
  print("emitting \(name) \(a)")
}
```

migueldeicaza:
Correct, it is wasteful, and would love to consolidate, that was filed as https://github.com/migueldeicaza/SwiftGodot/issues/42

I am open to the changes if we can harmonize them with the existing signal setup, would this change break existing code that might have used #signal already?

samdeane:
Better mockup - this code actually runs and the test passes, although it's obviously just a sketch...

```swift
import ChaosTesting
import Testing

@testable import Scratch

@Test func name() async throws {
  struct Foo {
    var foo: GenericSignal<Int> { GenericSignal<Int>(name: "foo") }
  }

  var callbackRan = false
  let f = Foo()
  f.foo.connect { (i: Int) in
    #expect(i == 42)
    callbackRan = true
  }

  f.foo.emit(42)
  #expect(callbackRan)
}

struct GenericSignal<each T> {
  let name: String

  func emit(_ t: repeat each T) {
    emitSignal(name, repeat each t)
  }

  func connect(callback: @escaping (_ t: repeat each T) -> Void) {
    sp.proxy = { args in
      callback(repeat args.pop(as: (each T).self)!)
    }
  }
}

nonisolated(unsafe) var sp = SignalProxy()

/// A packed array of arguments that can be popped off one by one,
/// with type checking.
class Arguments {
  init(_ args: [Any]) {
    self.args = args
  }

  var args: [Any] = []
  func pop<T>(as: T.Type) -> T? {
    let v = args.first as? T
    args.removeFirst()
    return v
  }
}

/// A proxy object that can be registered with a signal.
/// When the signal is emitted, the proxy will be called,
/// and will invoke its callback with the supplied arguments.
struct SignalProxy {
  var proxy: ((Arguments) -> Void)? = nil

  func proxyFunc(args: Arguments) {
    proxy?(args)
  }
}

func emitSignal<each T>(_ name: String, _ args: repeat each T) {
  var a: [Any] = []
  repeat a.append(each args)
  sp.proxyFunc(args: Arguments(a))
}
```

samdeane:
> would this change break existing code that might have used #signal already?

It might be possible to support some kind of backwards compatibility, but I suspect that a cleaner route could be to implement a new solution and then deprecate the old one.

I am wondering whether the new solution could use `@Signal` instead.

So you'd do something like:

```swift
@Godot class MyClass {
  @Signal var mySignal

```

and it would expand the body for you, out to:

```swift
@Godot class MyClass {
  var mySignal: GenericSignal< > { GenericSignal< >(target: self, signalName: "mySignal")
```

samdeane:
So you could also do:

```swift
@Godot class MyClass {
  @Signal(args: [String.self, Int.self]) var mySignal

```

and it would expand the body to:

```swift
@Godot class MyClass {
  var mySignal: GenericSignal<String, Int> { GenericSignal<String, Int>(target: self, signalName: "mySignal")
```

and so on...

samdeane:
I will write up the purpose of this PR as a new issue.

I think it would then be possible to implement #42 using my suggested solution (a non-breaking change), then go on to implement the new issue.

samdeane:
Note that:
- this builds on #587 so potentially it'd be easier to review & ideally merge that first
- this change is fully **additive** at the moment and leaves the old syntax alone
- there are additional bits of work that could be done - but I didn't want the scope of this PR to get out of hand
   - deprecate (or remove) the old `#signal` macro
   - switch to `@Signal` for the new macro?
   - update the Signal documentation -- I'm happy to do this but wanted to wait on a decision about deprecation etc

samdeane:
I've cleaned up the formatting - the PR should now just contain the actual changes I've made.

migueldeicaza:
It looks like the Windows build is broken:

```
2024-11-01T14:33:56.8202285Z D:\a\SwiftGodot\SwiftGodot\Sources\SwiftGodot\Core\GenericSignal.swift:49:20: error: value pack expansion can only appear inside a function argument list or tuple element
```

Maybe parameter Packs are not supported on Windows yet?

samdeane:
> Maybe parameter Packs are not supported on Windows yet?

Uh-oh. I'll take a look.

samdeane:
Closing this in favour of #626, which has a cleaned up description.

PR #587, opened 2024-10-24
GenericSignal helper class

samdeane:
See also #42 

Instead of generating a helper class for each signal property, we use a single GenericSignal which can be specialised with the list of argument types that the signal contains.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -41,5 +41,9 @@ jobs:
           tag: 6.0.2-RELEASE
       - uses: actions/checkout@v4
       - name: Build
-        run: swift build -v
+        run: swift build --build-tests --quiet
+      - name: Run tests (XCTest)
+        run: swift test --skip-build --no-parallel --disable-swift-testing
+      - name: Run tests (Swift testing)
+        run: swift test --skip-build --no-parallel --disable-xctest
         
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -37,8 +37,8 @@ jobs:
     steps:
       - uses: compnerd/gha-setup-swift@main
         with:
-          branch: swift-5.10-release
-          tag: 5.10-RELEASE
+          branch: swift-6.0.2-release
+          tag: 6.0.2-RELEASE
       - uses: actions/checkout@v4
       - name: Build
         run: swift build -v
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -34,6 +34,9 @@ jobs:
 
   build-windows:
     runs-on: windows-latest
+
+    timeout-minutes: 10
+    
     steps:
       - uses: compnerd/gha-setup-swift@main
         with:
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.9
+// swift-tools-version: 5.10
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
@@ -231,7 +231,7 @@ targets.append(contentsOf: [
 let package = Package(
     name: "SwiftGodot",
     platforms: [
-        .macOS(.v13),
+        .macOS(.v14),
         .iOS (.v15)
     ],
     products: products,
--- a/Sources/SwiftGodot/Core/GenericSignal.swift
+++ b/Sources/SwiftGodot/Core/GenericSignal.swift
@@ -27,7 +27,7 @@ public class GenericSignal<each T: VariantStorable> {
     /// - Returns: an object token that can be used to disconnect the object from the target on success, or the error produced by Godot.
     ///
     @discardableResult /* Signal1 */
-    public func connect(flags: Object.ConnectFlags = [], _ callback: @escaping (_ t: repeat (each T)?) -> Void) -> Object {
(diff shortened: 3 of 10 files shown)


Package.swift
samdeane, line 1: Apologies for the reformatting - I'm using swift-format automatically.

There is an actual change buried here, which is to bump up the macOS required version to v14. This seems to be needed in order to use the generic pack stuff.

samdeane:
Not sure the best way to comprehensively test this.

I think the one thing that might fail at runtime is the code that is unpacking the arguments to pass to the connect closure.

It is assuming that they can be constructed from a Variant. This should be true, but it's not the same code as was in the old generated helpers.

I've tested a bit with my own Godot project, but it would be good to try some others that make use of Godot-defined signals with parameters, and check that they work.

samdeane:
(also, I currently have some [completely unrelated](https://github.com/swiftlang/swift-package-manager/issues/7930) problems running the tests right now 🤦)

samdeane:
Any chance that someone could approve the workflow so that I can see if the tests pass?

cc/ @migueldeicaza @elijah-semyonov

Sources/SwiftGodot/Core/GenericSignal.swift
```diff
@@ -0,0 +1,147 @@
...
+extension Arguments {
+    enum UnpackError: Error {
+        case typeMismatch
+        case missingArgument
+    }
+
+    /// Unpack an argument as a specific type.
+    /// We throw a runtime error if the argument is not of the expected type,
+    /// or if there are not enough arguments to unpack.
+    func unpack<T: VariantStorable>(as type: T.Type, index: inout Int) throws -> T {
```
samdeane, line 124: I was hoping I could just call makeOrUnpack here, but the compiler can't decide between the two implementations and so says that they are ambiguous. 

I think that this is because T is an existential VariantStorable here, and so the compiler has lost any knowledge of whether it's also an Object.

Similarly I was hoping I could call `asObject(type)`, but the same problem exists - it complains that T needs to be an object.

I also tried specialising `func unpack<T: VariantStorable>(as type: T.Type, index: inout Int)` for T: Object, hoping that the pack expansion at the call site would pick the right version for each T -- but it doesn't seem to.

Someone smarter than me might be able to figure that stuff out, in which case this code could be a bit cleaner.

samdeane:
I've cleaned up the formatting - the PR should now just contain the actual changes I've made.

.github/workflows/swift.yml
```diff
@@ -37,8 +37,8 @@ jobs:
     steps:
       - uses: compnerd/gha-setup-swift@main
         with:
-          branch: swift-5.10-release
-          tag: 5.10-RELEASE
+          branch: swift-6.0.2-release
```
samdeane, line 40: @migueldeicaza I thought I'd try bumping the version right up to 6.0.2 here to see if the Windows compiler's argument pack support was fleshed out. The tests pass so I guess that it is.

I guess it might be worth setting the swift-tools-version to 6.0 if we truly do need Swift 6 on Windows - but then we'd have to add an explicit opt-out of the Swift 6 language model for now, to avoid having to deal with that whole can of worms...

Sources/SwiftGodot/Core/GenericSignal.swift
samdeane: Update: @migueldeicaza maybe @elijah-semyonov's variant changes will have made this easier; I haven't looked yet, but will try to.

samdeane:
Note that with @elijah-semyonov's variant changes, I think the generic class has to deal with the possibility of one of the arguments passed to a signal-with-argument could be nil. 

I'm not sure if that's actually possible in reality? Since this code was written before that change was merged, for now I've just taken the simplest route and I throw an error for a nil argument, just like I would for one of the wrong type. I can probably eventually get nil arguments to work fine if they are needed, but I wanted to minimise the changes after rebasing, so it might be simpler to tackle it in a follow-up PR later.

migueldeicaza:
The lack of support for `nil` is troubling, because there is a world where we would need to pass a nil that is being surfaced by Godot to our code.

One solution might be to bring back a `Variant.Nil` that conforms to `VariantStorable` so we can satisfy this signature:

```
    public func connect(flags: Object.ConnectFlags = [], _ callback: @escaping (_ t: repeat each T) -> Void) -> Object {
```

Or perhaps the only scenario we care about is passing an object that might be nil.

I think that we might be able to get away with creating such an object, and manually set the handle to nil, and users of that code would need to test for 'object.isValid'.

migueldeicaza:
I am not a fan of 'object.isValid' approach, so perhaps we do need to introduce a Nil type, perhaps we can call it 'CallbackNil' to make it clear that it is to be generally shunned.

Sources/SwiftGodot/Core/GenericSignal.swift
samdeane: I've cleaned this up a bit, and added an explanatory comment about why we can't just call T.unpack.
I'm happy with this now.

samdeane:
> The lack of support for nil is troubling

I've revised this so that nil arguments should be passed through to the callback properly.

samdeane:
After further discussion with @elijah-semyonov I've changed the generator as he suggested, so that primitive signal arguments are non-optional, but object ones are optional, in the actual signature of the GenericSignal.

migueldeicaza:
Thank you for nurturing this change.

Cheers!

PR #591, opened 2024-10-27
Fix build

sekimondre:
Somehow the build was broken. I just put a parenthesis back and it worked.

Can someone check it just to verify?


migueldeicaza:
Thank you!   I did not see this before I landed a similar patch, but thank you!

PR #593, opened 2024-10-28
Fixes @Callable macro expansion for Godot's collection params

sekimondre:
### Description

Fixes a build time error where expanding the `@Callable` macro in functions that have `VariantCollection<T>` or `ObjectCollection<T>` parameters calls for a missing `value:` label.

### Example Function
```swift
@Callable
    func importResource(
        sourceFile: String,
        savePath: String,
        options: GDictionary,
        platformVariants: VariantCollection<String>,
        genFiles: VariantCollection<String>
    ) -> Int {
    //...
}
````


elijah-semyonov:
I think we should omit the `value` label instead to follow Swift convention.

migueldeicaza:
The fix for `append` went in as part of the nil support, so this should be fixed now.

PR #595, opened 2024-10-29
Add swift-format config

samdeane:
Adds a `.swift-format` file to the root of the project.

If swift-format is used to format any files, it will pick up these settings.

Note that this won't magically cause all files to be formatted, and I've not run the formatter over them.

Having this file here just means that if someone *does* run swift-format, it will do something consistent and not use whatever their default settings are.

The actual settings we might want to settle on are of course up for grabs. I've picked some that are pretty close to Apple's defaults, but not identical.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,0 +1,69 @@
+{
+  "fileScopedDeclarationPrivacy": {
+    "accessLevel": "private"
+  },
+  "indentation": {
+    "spaces": 4
+  },
+  "indentConditionalCompilationBlocks": true,
+  "indentSwitchCaseLabels": false,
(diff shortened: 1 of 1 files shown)


samdeane:
I think the latest Xcode has support for running swift-format, and I believe that it would respect the settings in the file if it's there, but I haven't tried it. I will experiment a bit!

samdeane:
I've tested this. 

Xcode does honour the settings in the file, but only if you explicitly format a source file using Ctrl-Shift-I. 

image

So currently this change won't affect anyone who doesn't opt in, but it will set a baseline

Issue #585, opened 2024-10-24
DefaultValue support 


migueldeicaza:
We have two kinds of default values that are not yet supported:

```
Generator/default_value: no support for [Variant] = 0
Generator/default_value: no support for [typedarray::RDPipelineSpecializationConstant] = Array[RDPipelineSpecializationConstant]([])
```

Other than that, we are done.

PR #596, opened 2024-10-29
Suppress Generator warnings

samdeane:
Disabled the generator's `no support for...` warnings by default.

They can be enabled by defining WARN_MISSING, by adding `"-Xswiftc" "-D" "-Xswiftc" "WARN_MISSING"` to the build arguments when building with SPM (or doing the equivalent in Xcode).


migueldeicaza:
Thanks!   I had forgotten to remove those when I filed the companion bug:

https://github.com/migueldeicaza/SwiftGodot/issues/585

PR #597, opened 2024-10-29
Explicitly exclude files to suppress warning.

samdeane:
When building, the following warning is emitted:

```
warning: 'swiftgodot': found 2 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /Users/sam/Developer/Projects/Godot/Sandbox/SwiftGodot/Sources/ExtensionApi/ApiJsonModel.swift
    /Users/sam/Developer/Projects/Godot/Sandbox/SwiftGodot/Sources/ExtensionApi/ApiJsonModel+Extra.swift
```

This change does what the warning suggests!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,8 +1,8 @@
 // swift-tools-version: 5.9
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
-import PackageDescription
 import CompilerPluginSupport
+import PackageDescription
 
 //var linkerSettings: [LinkerSetting] = []
 //#if os(macOS)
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thank you for the patch, I have manually applied the change, there was a lot of formatting changes that came in through - I do not mind fixing those, but that needs to be part of a separate change.

samdeane:
Yeah, sorry, that'll be me using swift-format again. Much more sensible to just apply the patch.

I have vscode set to run swift-format on save, which is normally great but not so good whilst I'm the only one doing it!

PR #598, opened 2024-10-29
Fail the GH build/test steps if the swift build/test actually failed.

samdeane:
Currently, if the `swift build` or `swift test` commands fail, the GH action step will still be marked as succeeded.

This seems to be an artefact of the way we were re-running the tests from within the Godot engine; it resulted in the overall failure count being 0 even if there were actual failures.

This PR changes the way we run the tests, and should result in a non-zero exit status if a test does fail.

It also separates the XCTest run and the Swift testing run into two steps.

The default behaviour of `swift test` is to first run the XCTest engine, then run Swift testing engine. 

This results in the following confusing output at the end of the test run:

```
􀟈  Test run started.
􀄵  Testing Library Version: 102 (arm64e-apple-macos13.0)
􁁛  Test run with 0 tests passed after 0.001 seconds.
```

This is the output from the Swift testing engine. Because all our tests are using XCTest, and we have no Swift testing tests, it is saying that 0 tests passed. As it's the last thing in the report, it's a bit confusing and for a while I thought that this was why the exit status was 0.

To avoid this confusion, we can explicitly choose a test ending by passing `--disable-swift-testing` or `--disable-xctest`.

I have done this in two separate steps, so that the output from each testing engine is clearer. 

The Swift testing engine step is a null-op right now, so we could remove it, but at some point we might want to migrate some tests to the new testing framework, so I figured it was worth having it there to avoid later confusion.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -29,6 +29,11 @@ jobs:
       run: swift build -v
     - name: Run tests
       run: set -o pipefail && swift test -v --xunit-output test-results.xml
+    - name: Upload test results
+      uses: actions/upload-artifact@v4
+      with:
+        name: test-results
+        path: test-results.xml
   build-windows:
     runs-on: windows-latest
     steps:
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -25,14 +25,12 @@ jobs:
       uses: maxim-lobanov/setup-xcode@v1
       with:
         xcode-version: "16.0"
-    - name: Install XCPretty
-      run: gem install xcpretty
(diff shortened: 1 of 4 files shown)


samdeane:
Update: `swift test` seems to return 0 even when there's a failure... which is weird.

samdeane:
I wonder if we can use the `--xunit-output` then run a report over the results?

samdeane:
Eg with something like
- https://github.com/marketplace/actions/junit-report-action
- https://github.com/marketplace/actions/test-reporter

migueldeicaza:
Thank you so much!

PR #599, opened 2024-10-31
Generator Output List Sorted

samdeane:
I was seeing a weird thing where every time I ran build, the generator would run again, even if I'd just built.

I started wondering whether the output list being pushed in to the buildCommand was wrong. If that contained a file that didn't exist, it would explain the generator running each time.

So I pasted in the generated file names into to the source code and cleaned them up - and it seems to have fixed the problem.

The odd thing is that as far as I can see, I haven't changed any file names, I've just sorted them. I happened to take the names from the Finder rather than the output of ls, which has changed the order slightly. 

Maybe there's a bug in the SPM plugin code relating to the sort order?
Maybe this is a complete false-positive?


samdeane:
Me too. According to @heckj [it may just be a "feature" of SPM at the moment](https://swiftongodot.slack.com/archives/C05BYJ95S5D/p1730386855136869?thread_ts=1730381161.714269&cid=C05BYJ95S5D)

PR #604, opened 2024-11-14
Refactor Extension Interface

konczg:
These changes refactor the extension interface into a single protocol that provides access to the library handle and the getProcAddr function, giving the implementation more flexibility in providing these to the bindings.


migueldeicaza:
This broke the build, but the other problem is that it introduced an "import GDExtension" when we use "@_implementationOnly import GDExtension" everywhere else.

konczg:
I have recreated the PR with the necessary changes: https://github.com/migueldeicaza/SwiftGodot/pull/611

PR #612, opened 2024-11-18
Add Swift 6 docs

DandyLyons:


DandyLyons:
One of the Swift contributors suspects that we might have found a compiler bug. We should hold off on adding these docs until we've investigated further. 

https://forums.swift.org/t/crowd-source-swift-language-modes-in-various-environments/76102/3

DandyLyons:
I found the problem thanks to Xiaodi. https://forums.swift.org/t/crowd-source-swift-language-modes-in-various-environments/76102/6

Apparently the default language mode is Swift 6 if you use `// swift-tools-version: 6.0` in your `Package.swift`. 

I'll update this PR to reflect that later.

migueldeicaza:
There is another PR #615 that fixes the build for 6.0, does the documentation here need to be updated, or would it be good to go once we land the other one?

DandyLyons:
Currently users can use either Swift 6 tooling or Swift 5 tooling. After PR #615 we would only be able to use Swift 6 tooling, so a lot (probably most) of these new docs in this PR would be irrelevant. However, users would still be able to choose which language mode they want to use for each target. 

I'd say once PR #615 is finalized, we can assess what changes this PR needs to be accurate.

DandyLyons:
The tl;dr; is that after we require Swift 6 tooling, (after PR #615), then each module in a package will be Swift 6 language mode by default. This could be a very unexpected change, and so the docs should advise people to opt into Swift 5 language mode, and tell them they can opt into Swift 6 when they are ready. 

I explored this in greater detail here: https://dandylyons.net/posts/post-26/am-i-using-swift-5-or-6/

Even more confusingly, that is the behavior if SwiftGodot is added as a dependency directly in a Package.swift file. But the behavior is not the same if a package dependency is added through the Xcode GUI. 

It's also important to note that Matt Massicote advises that we should NOT use Swift 6 language mode until we have a compelling reason to. (Of course, his opinion isn't infallible, but when he advises on Swift Concurrency, I pay attention. He wrote a huge portion of the official Swift 6 transition guide.)
https://www.youtube.com/watch?v=oMg2OjO3LLI

PR #624, opened 2024-12-05
Fix filtered tests

samdeane:
This PR rewrites the test harness to allow individual tests to be run again, as well as running the full suite. If a test fails, the testing process will correctly exit with a non-zero status.

You can run individual tests:
- using the individual run test buttons in Xcode image
- using the individual run test buttons in VSCode image
- using the command line: `swift test --filter=myTestPatternHere`

The approach used to run the tests is described [here](https://github.com/elegantchaos/EmbeddedXCTest).
It is similar to before, but uses a slightly different approach which preserves the filtered list of tests.


migueldeicaza:
Delightful!   Thank you so much!

Issue #586, opened 2024-10-24
Treat Signals Declared In Swift Like Signals Declared In Godot


samdeane:
When we import a signal from Godot, we generate a helper class and a computed property, which allows us to connect to it in a very Swift-like manner:

(insert example here)

When we define a signal in Swift using the `#signal` macro, we create a static property on the Swift class. This requires a different, less intuitive, syntax for connected to or emitting signals.

```
let actor = SBActor()
actor.connect(signal: SBActor.weaponsChanged, to: self, method: "updateWeapons")
```

It should be possible to re-write the signal support so that Swift-side signals are implemented using the same helper classes as imported ones. This should result in a more uniform interface, and less code.

This could be a breaking change (if we alter the `#signal` macro), but I think it could probably be done instead as an additive change, using a different macro (possibly `@Signal`). Any implementation would probably be built on top of #42.

samdeane:
A solution would result in syntax more like this:

```swift
@Godot public class SBActor: Object {
    @Signal var weaponsChanged   // signal will be exposed to Godot as "weapons_changed"
    
  func changeWeapon(_ weapon: SBWeapon) {
    weaponsChanged.emit()
  }
}

@Godot public class SBHud: Object {
  func connectToActor(_ actor: SBActor) {
    actor.weaponsChanged.connect {
      // update the HUD UI here...
  }
}

**Comment by samdeane** [comment]
It should also support signal parameters properly:

```swift
@Godot public class SBActor: Object {
  @Signal(args: [Int.self]) var healthChanged   // signal will be exposed to Godot as "health_changed, <new-health>"
    
  func hit(_ amount: Int) {
    health -= amount
    healthChanged.emit(health)
  }
}

@Godot public class SBHud: Object {
  func connectToActor(_ actor: SBActor) {
    actor.healthChanged.connect { health in
      // update the HUD UI here...
  }
}
```

PR #626, opened 2024-12-06
Add @Signal Macro

samdeane:
Implements #586.
See also previous discussion on #584.

## TL;DR

This PR adds a new `@Signal` macro, giving a cleaner and more consistent way to add user-defined signals from Swift. 

```swift

@Godot
class MyNode: Node {
  @Signal var mySignal: SimpleSignal
  @Signal var signalWithArgs: SignalWithArguments<Int, String>
}
```

These signals can be used in the same way as built-in signals:

```swift
node.mySignal.connect { /* do something */ }

node.signalWithArgs.connect { intArg, stringArg in
  // do something
}

node.mySignal.emit()
node.signalWithArgs.emit(1, "foo")
```

## Motivation

The existing `#signal` macro defines the signal as a **static** property on the class that contains it.

This is different from the generated implementation of built-in signals, and means that we have to use a different syntax to work with them -- which feels... non-optimal.

## Implementation

We add a new `@Signal` macro expansion which generates a computed property on the class, using `GenericSignal`. 
This is the same approach that is used to implement the signals imported from the Godot API.

We also add an `emit()` method to `GenericSignal`, which is of course essential to support emitting signals defined in Swift.

We add `SimpleSignal` and `SignalWithArguments` as slightly clearer aliases for `GenericSignal`

This change should be 100% additive, and should not break the old `#signal` macro.

## To Do

If this change is adopted, it's probably worth updating the documentation and deprecating the old macro.


Generator/Generator/MethodGen.swift
```diff
@@ -30,6 +30,10 @@ enum GeneratedMethodKind {
     case utilityFunction
 }
 
+/// The list of static method names to make public.
+/// Most of the names are kept fileprivate.
+let publicMethodNames = ["method_get_class", "method_emit_signal"]
```
samdeane, line 35: I had to make this public in order to implement an alternative version of `emitSignal`, which the `GenericSignal` implementation needs in order to implement an `emit()` method with the correct arguments. See also `RawCall.swift`.

I suspect that there may be a better way. Possibly I can use a Variant to make the call?

Sources/SwiftGodot/Core/RawCall.swift
```diff
@@ -0,0 +1,45 @@
+@_implementationOnly import GDExtension
+
+extension Object {
+
+    /// Make a raw call to a method bind.
+    /// All input arguments must be marshaled into `Variant`s.
+    /// The result is a `Variant` that must be unmarshaled into the expected type.
+    @discardableResult /* discardable per discardableList: Object, emit_signal */
+    final func rawCall(_ p_method_bind: GDExtensionMethodBindPtr, arguments: [Variant]) -> Variant? {
```
samdeane, line 9: I can't help feeling that there's a better way to do this!
Calls to the API are all generated, with the object handle as the first parameter, and the other arguments as a variadic parameter.

I needed to instead make a call that takes an array of arguments.

migueldeicaza:
This is lovely!   Would you mind testing the Object.callv approach?  If it works, that might be better to maintain in the long run.

samdeane:
Using `Object.callv` seems to work well.
Have also squashed all the commits for a cleaner history.

migueldeicaza:
Thank you!

Merging.

PR #633, opened 2024-12-18
Split Generated Source Into 26 files for Windows

samdeane:
[**NOTE**: this PR includes the changes from #615, so it would be easier to review once that one is merged]

## The Problem

Windows is failing to build on the CI machine.

It appears to be due to the humungous size of the combined generated file, which I suspect causes some sort of exponential explosion in the compiler; the upshot of which is that it just hangs.

This PR attempts to fix that.

## The Solution

Instead of generating a single enormous file, we generate 26 files `SwiftGodotA.swift`, `SwiftGodotB.swift` etc.

Each file contains all the types that start with that letter (or more specifically, all the types output by Printer instances whose names start with that letter).

This keeps the number of input files down to a manageable amount, but also the size of each file from getting too large.

If necessary we could use a different hash to produce more, smaller, files, but this seems to be enough to get the build to actually finish.


samdeane:
I still takes ~20 mins to build on Windows, but it does at least build now, and the minimal tests that don't need libgodot do actually link & run.

migueldeicaza:
Could we undo the whitespace changes on the files that are unrelated to this?   Like the EntryPoint changes have a lot of churn.   

I am trying to merge a large set of internal changes for object ownership, and the whitespace changes every time become very time consuming.

migueldeicaza:
Question, the generated split only applies to Windows, but on Unix we still keep all of our files the way we have them now?

samdeane:
>  Question, the generated split only applies to Windows, but on Unix we still keep all of our files the way we have them now?

That's the intention, yes.

samdeane:
Does Linux have a similar problem with path lengths & the linker?

I was actually thinking that I might try to add a Linux runner to the workflow (later, not as part of the PR) - but I wasn't sure what state the linux build was in.

migueldeicaza:
I am not aware of issues with Linux, I think that would be fine.

migueldeicaza:
The original comment states that I should merge firs the #615 change, but it seems like that one has the build broken, which this one fixes.

Should I just merge this one instead?

migueldeicaza:
Btw, it would be really nice if we did not include whitespace changes on these PRs.

samdeane:
> Btw, it would be really nice if we did not include whitespace changes on these PRs.

I only ever do it by accident. 

I try to remember to turn off automatic format-on-save now, but I still forget sometimes, especially if I'm editing the project as a submodule inside a larger workspace.

To be honest this is likely to continue to be a problem unless we come up with some formatting rules and a way to automatically apply them. I don't really care what the formatting rules actually are - I just want to be able to run a single command to fix things when I make a PR!

Of necessity there will be some initial pain with this for people who have a large patch/branch they're trying to maintain (like you are?) but I don't see any way round it. Even then if you could apply the same formatter to both your patch and the target branch, I think the diff would be clean(er?).

samdeane:
Added #646 to track the idea of automating formatting.

PR #634, opened 2024-12-19
Run Macro Tests On Windows

samdeane:
Removes @testable from all unit tests.
This should prevent the exported symbol count from leaping through the roof when building SwiftGodot for testing.
This in turn should allow us to run the macro tests on Windows.


Package.swift
```diff
@@ -121,7 +121,14 @@ var targets: [Target] = [
             .product(name: "SwiftSyntax", package: "swift-syntax"),
             .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
         ],
-        swiftSettings: [.swiftLanguageMode(.v5)]
+        swiftSettings: [
+            .swiftLanguageMode(.v5),
+            .unsafeFlags(
+                [
```
samdeane, line 127: Without this, the compiler was complaining about there being a main() entrypoint in this target which clashes

Sources/SwiftGodot/EntryPoint.swift
```diff
@@ -7,62 +7,6 @@
 
 @_implementationOnly import GDExtension
 
-public protocol ExtensionInterface {
```
samdeane, line 10: I moved this stuff to its own file for clarity.
I think the stuff in here and the new file could do with cleaning up & refactoring, but that can wait for another PR...

PR #649, opened 2025-01-17
Deprecate old signal API

samdeane:
If we're happy with the new `@Signal` macro, we should probably aim to strip the old one out eventually.

A first step is to deprecate the old one, and the associated API.


samdeane:
Unit test failure seems to be in `WrappedTests.testRetain` which is unrelated to this PR as far as I can see.

image

migueldeicaza:
Thank you Sam!

PR #650, opened 2025-01-22
Extend convenience initializer coverage for Packed*Array

glui:
Noticed that there weren't any straightforward ways to construct PackedVector\*Arrays from arrays of Vector\*, so I hacked this together. It appears to work and passes tests. Thank you for the awesome project!

**Covered:**
- PackedColorArray
- PackedVector2Array
- PackedVector3Array
- PackedVector4Array

**Not Covered:**
- PackedStringArray


migueldeicaza:

PR #652, opened 2025-01-23
Build As DLL, Use Swift 6.1 prelease, on Windows

samdeane:
Contains the following changes:

- brings Windows into line with the other platforms by building SwiftGodot as a shared library
- updates the Xcode/Swift versions in the CI workflow
- removes a workaround for the Generator plugin on Windows, which no longer seems to be necessary when building with Swift 6.1.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -40,8 +40,8 @@ jobs:
       - name: Swift Install
         uses: compnerd/gha-setup-swift@main
         with:
-          branch: swift-6.0.3-release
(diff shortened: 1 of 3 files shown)


samdeane:
I guess that, if it does indeed fail to build on 6.0.3, we could conditionalise the fix.
That said, I also had problems with the plugin failing on 6.0.3, so I wonder if this is a fix that will work for both. The CI should tell us.

samdeane:
The CI did fail on 6.0.3 - trying with an actual 6.1 snapshot instead.

PR #653, opened 2025-01-24
Fix Xcode version for CI

samdeane:
Github have updated their Xcode versions, and no longer have 16.0, so we need to update to 16.2.
At the same time, upgrading the Windows CI to a pre-release of 16.1 potentially fixes some problems with the generator plugin.


samdeane:
Note I pulled these two out of #652 in case you just want to take them on their own first.

PR #658, opened 2025-01-30
Cleaner Example Extensions

samdeane:
Cleans up SimpleExtension a little.
Adopts the EntrypointGenerator plugin, and removes the example of manual registration which is not really "simple".

Adds a ManualExtension example with the code removed from SimpleExtension, illustrating how to:
- manually register properties etc
- manually write an entrypoint function


samdeane:
On a related note, is the Testbed Godot project supposed to work with SimpleExtension?

It looks like it's long out of date, and is currently broken.

I wonder if it would be better to make a completely new repo, and move Testbed, SimpleExtension and/or ManualExtension to it, then direct people to that project for an example.

migueldeicaza:
The testbed was just for my own use for quick API tests, happy to fix it.

We could also move the samples elsewhere, I am just afraid they will be forgotten and go unmaintained.   Keeping them on the repo at least ensures that they are built along with SwiftGodot (and I agree they are in bad shape as far as being a source of sample material)

PR #659, opened 2025-01-30
@Node macro

samdeane:
I don't love the name or ergonomics of the @SceneTree macro.

The `path:` label for the argument is visual noise.

Requiring the variable type to be optional or implicitly unwrapped also doesn't quite make sense - we may as well support it being non-optional. For non optional variables we'll get a runtime error if the node is missing, which is also true for implicitly unwrapped vars.

This PR introduces a @Node macro. It uses the same macro implementation as @SceneTree, but tweaks the interface to be:

```swift
@Node("nodePath") var myNode: SomeNodeType
@Node("nodePath") var myOptionalNode: SomeNodeType?
```

The scene tree macro implementation has been tweaked to support non-optional type definitions, rather than complaining about them. It does still support implicitly unwrapped types, for backwards compatibility with @SceneTree, which remains unchanged.

### DIFF
--- a/Sources/SwiftGodot/MacroDefs.swift
+++ b/Sources/SwiftGodot/MacroDefs.swift
@@ -216,6 +216,24 @@ public macro NativeHandleDiscarding() = #externalMacro(module: "SwiftGodotMacroL
 @attached(accessor)
 public macro SceneTree(path: String? = nil) = #externalMacro(module: "SwiftGodotMacroLibrary", type: "SceneTreeMacro")
 
+/// A macro that finds and assigns a node from the scene tree to a stored property.
+///
+/// Use this to quickly assign a stored property to a node in the scene tree.
+/// ```swift
+/// class MyNode: Node2D {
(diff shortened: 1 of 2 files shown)


migueldeicaza:
We should get it documented.   I believe the docs might still point to the original version that might have used property wrappers and was very flaky.   The one that came before SceneTree

samdeane:
I've updated the docs and examples to use `@Node` instead of `@SceneTree`.

Also removed the references to the old `BindNode`.

migueldeicaza:
I think we should deprecate BindNode, it relies on some runtime reflection, which turns out breaks on some platforms as the reflection produces different (and non-deterministic) output.

migueldeicaza:
Just to confirm: old code using SceneTree will continue to work right?

(I do not use it myself, but I rather not break people's code).

samdeane:
It should do, yes. The intention was to add a new macro with more compact syntax, and to leave the old one alone.

PR #660, opened 2025-01-30
Remove Generated Enum debugDescription

samdeane:
See [this conversation](https://swiftongodot.slack.com/archives/C05BYJ95S5D/p1737984087090499).

We're generating a debugDescription for enums, but it's not really required.
Removing it will lower the exported symbol count, which is a GoodThing™

### DIFF
--- a/Generator/Generator/Enums.swift
+++ b/Generator/Generator/Enums.swift
@@ -90,7 +90,7 @@ func generateEnums (_ p: Printer, cdef: JClassInfo?, values: [JGodotGlobalEnumEl
         }
         let extraConformances = enumDefName == "Error" ? ", Error" : ""
         
-        p ("public enum \(getGodotType (SimpleType (type: enumDefName))): Int64, CaseIterable, CustomDebugStringConvertible\(extraConformances)") {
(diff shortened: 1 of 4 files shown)


migueldeicaza:
Conversation for the record:

> In other news: I noticed that the generator for enums adds CustomDebugStringConvertible conformance and defines a debugDescription property which is just the case name with a . prepended. In the quest for exported-symbol reduction, this seems like something we could do without, given that there are quite a few of them. Any thoughts?
> presumably each one of these calculated properties will result in an exported symbol
[8:25](https://swiftongodot.slack.com/archives/C05BYJ95S5D/p1737984339794219?thread_ts=1737984087.090499&cid=C05BYJ95S5D)
incidentally I looked into whether we could replace these with a single implementation in an extension, but sadly there is no way to define a debugDescription method which calls on to the default implementation :disappointed:
>  You only need to define a custom debugDescription if you want to override the default behaviour. In this case the only difference would be losing the initial “.”
> String(describing: x) will work for any enum case. Likewise “\(x)” in a string. Defining description or debugDescription should only be needed if you want to move away from that:

PR #663, opened 2025-01-31
New memory management system for SwiftGodot for Godot Objects.

migueldeicaza:
In this change we deal with proper ownership of objects in Godot, the RefCounted objects that Godot releases via ref/unref operations, and those that are queued for destruction.

To free objects:

* If you have a Node subclass, use `queueFree`
* If you have a RefCounted object, just clear your Swift Reference
* Otherwise, call `free`

In this change, we take advantage of Godot's system that notifies us when a Godot object has been released so we clear our "handle", and the `isValid` method that we have had for a while is now correct on all cases.

It is important because otherwise we could have a situation where Godot frees an object, but we still keep a pointer to it.  And due to how Godot's object allocation works, as Godot created new objects, rather than a crash to an invalid memory address, we would just have a pointer to the wrong object and things would seem to work.

To implement this, we also had to change an old SwiftGodot design principle - in the past, we did not care if multiple Swift peers were created for the underlying object, because we use to consider them purely peers that could be recreated on demand.  But now, when we take advantage of Godot notifying us of the object being deallocated, we would only deallocate one instance, not all the instances.  So to fix this, we now only surface one SwiftGodot object wrapper for a given Godot Object.

Some special handling is done for different objects in the Godot hierarchy.  The way we now handle the lifecycle of Godot objects in Swift is the following: If an object is not a RefCounted, we keep a strong reference to it, so that only a call to the appropriate free method can delete the object.

If an object is a RefCounted, we keep track of the reference()/unreference() calls to it, and we keep a weak reference to it if there are no references to it in Godot, otherwise we keep a strong reference to it. This guarantees that the object can still be deleted when no reference to it exists anymore, while making sure that subtyped objects keep their state on the Swift side throughout the Godot object's lifetime. As the Swift wrapper also increases the reference count of the RefCounted, this means that once a RefCounted is passed to the Swift side, that Godot object will always be destroyed by the Swift wrapper eventually.

This also introduces a cute method called
`attemptToUseObjectFreedByGodot`, so that if you see it on a stack trace, it is a clear indicator that this took place.  This can happen on code that kept references to long-term held objects, in those cases, you must call `isValid` on the object.

The system now can toggle if we need to keep a strong reference to the underlying object via the WrappedReference class, which can strongify or weakify the changes.

While I did the initial work on making the objects be unique and clearing of the handle, the real hard work and the correct tracking of object ownership was done by Gabor Koncz and Gergely Kis at Migeran who worked on all the various and difficult corner cases.

In particular the new object lifecycle is tracked like this:

- Every time godot returns a RefCounted object in a Ref<> wrapper for a ptrcall, its reference count is incremeneted. As object identity results in the return of the same Swift proxy for the same RefCounted object, this means that every subsequent return should result in an unreference() call, so that the existence of the Swift proxy object always results in a single increment of the reference count.

- On the other hand, if a RefCounted object was returned through a non-RefCounted static return type (e.g. as an Object), then Godot did not increment its reference count. This means that on the first return of such an object, the reference count should be incremented by a reference() call, so that similarly the existence of the Swift proxy object always results in a single increment of the reference count.

- The ownsRef parameter is true iff Godot can pass ownership of a Ref<> wrapper to SwiftGodot, e.g. with a Ref<> return value of a ptrcall.

This work was the result of using a large SwiftGodot-based application on production through real life testing, where we slowly uncovered the original design flaws and worked to fix each of those one by one. That is why this commit is a bundle of fixes, as the original history contains various attempts at solving the problem until the real solution emerged.


Sources/SwiftGodot/Variant.swift
```diff
@@ -313,8 +330,31 @@ public class Variant: Hashable, Equatable, CustomDebugStringConvertible {
...
+            var newContent: ContentType = Variant.zero
+            var copyIndex = index
+            var valid: GDExtensionBool = 0
+
+            gi.variant_get(&content, &copyIndex.content, &newContent, &valid)
+            if valid != 0 {
+                if newContent == Variant.zero {
+                    return nil
+                }
+                return Variant(takingOver: newContent)
```
elijah-semyonov, line 346: This constructor is already failable and will return nil if the content is zero. This check is redundant

migueldeicaza: Oh thanks for spotting this!

migueldeicaza:
One issue we are having with the tests is that the disposing of RefCounted object needs to happen with a callDeferred, and as such this means that we need to trigger the main loop execution before asserting that a resource was not freed in test_513_leak2

migueldeicaza:
This patch has another problem, it no longer caches the result of methods:

```
    fileprivate static var method_finish: GDExtensionMethodBindPtr {
        let methodName = StringName("finish")
        return withUnsafePointer(to: &AESContext.godotClassName.content) { classPtr in
            withUnsafePointer(to: &methodName.content) { mnamePtr in
                gi.classdb_get_method_bind(classPtr, mnamePtr, 3218959716)!
            }

        }
        
    }
```

migueldeicaza:
Without the Callable optimization:

```
            // 1..770: 104
            // 771..2818: 32,872 (2047 new values)
            // 2819..6914: 98,408 (4095 new values) - so 16
            // 6915..15106: 229,480 (8191 new values) - so 16 bytes per value
            // 15107: 491,624..
            // 200000: 4 meg leak
            // 2000000

```

migueldeicaza:
The test suite is crashing during the cleanup, when it attempts to probe for information for the `TestNode` that is allocated in `MacroCallableIntegrationTests.testCountMixed`

PR #675, opened 2025-03-09
Surface _validate_property method in Godot for allowing custom properties to be controlled

migueldeicaza:
…


migueldeicaza:
The last commit crashes immediately:

>  With the latest commit, Godot crashes immediately after the first call to my validateProperty. If I comment out validateProperty again it’s fine, just to rule out any other change.

PR #679, opened 2025-04-08
Introduce VariantConvertible. Make Export macro use it

elijah-semyonov:
1. Move macro expansions to resource files, add little infra to dump latest expected expansions there when we change macro code.
2. Introduce `VariantConvertible`, the type I promised long ago and which I want to supersede our current `VariantStorable` and  `VariantRepresentable`.
3. Minor tweaks and clean ups.
4. Make `Variant` final.
5. Make `@Export` macro rely on static dispatch information instead of raw type comparison.
6. Introduce `_GodotBridgeable`. It's an internal API that will allow to move `PropInfo` generation logic to the type and context, allowing to free macros of type-checking altogether and just use the literal typenames (where required, such as function arguments) and identifiers from the code, instead of trying to reason about what they are on macro level, where the information is not enough.


Generator/Generator/BuiltinGen.swift
```diff
@@ -813,6 +812,14 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
             generateBuiltinMethods(p, bc, bc.methods ?? [], typeName, typeEnum, isStruct: kind == .isStruct)
             generateBuiltinOperators (p, bc, typeName: typeName)
             generateBuiltinConstants (p, bc, typeName: typeName)
+                        
+            p("public func toVariant() -> Variant") {
```
migueldeicaza, line 816: For new public API entry points, could we include the relevant documentation?

I think this would do:

/// Wraps this type into a Variant

Generator/Generator/BuiltinGen.swift
```diff
@@ -813,6 +812,14 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
...
+                p("Variant(self)")
+            }
+            
+            p("public static func fromVariant(_ variant: Variant) -> Self?") {
```
migueldeicaza, line 820: Similar for this one, we should generate:

/// Attempts to turn the Variant into this type.

Ideally, if we have the name of the type, we would use that name here and replace it in the docs.

Generator/Generator/ClassGen.swift
```diff
@@ -622,6 +625,34 @@ func processClass (cdef: JGodotExtensionAPIClass, outputDir: String?) async {
             virtuals = generateMethods (p, cdef: cdef, methods: methods, usedMethods: referencedMethods, asSingleton: asSingleton)
         }
         
+        if inherits == objectInherits {
+            p("public func toVariant() -> Variant") {
```
migueldeicaza, line 629: We should also generate documentation for thse methods.

Sources/SwiftGodot/Core/VariantConvertible.swift
```diff
@@ -0,0 +1,127 @@
...
+    func toVariant() -> Variant
+    
+    /// Extract ``Self`` from a ``Variant``. Returns `nil` if it's not possible. E.g. another type is stored in the `variant`
+    static func fromVariant(_ variant: Variant) -> Self?
+}
+
+/// Internal API. Interface for types that contains details on how it interacts with C GDExtension API.
+/// This is more specialized version of ``VariantConvertible`` for cases where an ability to be converted to and from a ``Variant`` is not enough.
+/// At the same time it allows ``VariantConvertible`` to be implemented by the user for coding arbitary values inside the ``Variant``
+public protocol _GodotBridgeable: VariantConvertible {
```
migueldeicaza, line 21: Would love to include in the documentation here the reason for the cases where it needs additional support, as I do not understand it myself now, and would love to have this information in the future.

Sources/SwiftGodot/MacroIntegration/MacroExportSet.swift
```diff
@@ -0,0 +1,279 @@
+//
+//  _macroExportSet.swift
+//  SwiftGodot
+//
+//  Created by Elijah Semyonov on 09/04/2025.
+//
+
+/// Internal API.  Builtin types.
```
migueldeicaza, line 8: I think it is fine to have APIs that are flagged as internal, but we should describe what they are used for in these docs.

migueldeicaza (changes requested):
I love the feature and the extraction of the tests into separate files.

I left some notes on missing public API docs, but in addition to that, I think that we should update the relevant documentation for places where things change, I am thinking of Variant.md.

Sources/SwiftGodot/MacroIntegration/MacroExportSet.swift
elijah-semyonov: It's inevitable in our case since macro can't call `internal` code

elijah-semyonov:
@migueldeicaza 
I've walked through your feedback

migueldeicaza:
Woohoo!   Merged!

PR #680, opened 2025-04-10
Iteration on _GodotBridgeable

elijah-semyonov:
1. All macro logic related to `@Export` relies on static dispatch instead of syntax analysis.
2. `.enum` and `.nodeType` are filled automatically for `@Export` when eligible.
3. No explicit type information is needed for `@Export` anymore.


elijah-semyonov:
Superseded by another branch

PR #684, opened 2025-04-17
SVG logo resource and updated README.md

cathyceit:
I've seen that someone from the community made this logo, but apparently there is no SVG resource for that!
I've recreated it so that everyone can enjoy it!


migueldeicaza:
Oh this is lovely!   the original logo got lost to the time, and we had been passing around an ugly scaled PNG.

Thank you!

PR #687, opened 2025-04-17
Adjust rawvalue init of GDExtensionVariantType to use Int32 instead o…

houtzager:
Fixes 

```
S:\swift\.build\checkouts\SwiftGodot\Sources\SwiftGodot\Variant.swift:282:63: error: cannot convert value of type 'UInt32' to expected argument type 'Int32'      
280 |         var newContent: ContentType = Variant.zero

281 |         var err = GDExtensionCallError()

282 |         gi.variant_construct(GDExtensionVariantType(rawValue: UInt32(type.rawValue)), &newContent, nil, 0, &err)

    |                                                               `- error: cannot convert value of type 'UInt32' to expected argument type 'Int32'
283 |         if err.error != GDEXTENSION_CALL_OK {

284 |             return .failure(toCallErrorType(err.error))
```


houtzager:
Been too slow.

PR #694, opened 2025-04-25
Add `autoSnakeCase` to @Callable

elijah-semyonov:
Add `autoSnakeCase` parameter to `@Callable`. It will allow customising how function name is exported to Godot.
Swift common convention is `camelCase` while Godot uses `snake_case` for functions.
Default value is `false`. In the future I expect to make it depend on SPM trait, so that default value is `true` for new projects, and `false` for projects that opt-in the legacy behavior.
That behavior will align with `@Export` macro, which I believe is the most convenient one.


migueldeicaza:
Oh this is great!   Thank you!

PR #697, opened 2025-04-26
Single `required init` chain for `Object`s.

elijah-semyonov:
### Single `required init` chain

There is only one `required init` chain now, that takes `InitContext`. In case user _do_ need a custom `init` they will have only:

```
    required init(_ initContext: InitContext) {
        super.init(initContext)
        
        addPoint (id: Self.A, position: Vector3 (x: 0, y: 0, z: 0))
        addPoint (id: Self.B, position: Vector3 (x: 1, y: 0, z: 0))
        addPoint (id: Self.C, position: Vector3 (x: 0, y: 1, z: 0))
        addPoint (id: Self.X, position: Vector3 (x: 0, y: 0, z: 1))
        connectPoints (id: Self.A, toId: Self.B)
        connectPoints (id: Self.A, toId: Self.C)
        connectPoints (id: Self.B, toId: Self.C)
        connectPoints (id: Self.X, toId: Self.A)
    }
 
```

They can't construct their own, it's required for `super.init`, so it's almost impossible to get wrong.

At the same time convenience initialiser `init()` is accessible for all `Object`-derived types, it will go through the same path but supply the newly constructed native object.

### `Identifiable` by design

`Object.id` for `Identifiable` conformance is removed as redundant. Before memory model rework we could have multiple framework types instances in Swift wrapping a single native instance. Now it's impossible and hence default `Identifiable` implementation of `AnyObject` is reflecting the current identity semantics.

### `Object.handle` removed from public API
Shouldn't be there.

### Internal improvements
Multiple renames, typealiases for what `UnsafeMutableRawPointer` actually means in internal implementation to make it more approachable.

Opaque `Object *` is kept as `UnsafeMutableRawPointer` instead of `UnsafeRawPointer` to align with the GDExtension API. No more awkward `UnsafeMutableRawPointer(mutating: constRawPointer)`.

### Singleton path
Godot singletons now get a lazy Swift bound object instead of Swift counterpart being `init`/`deinit` on every `shared` access


elijah-semyonov:
I'll work more thoroughly on it. I think there is a chance to ditch `WrappedReference`

Generator/Generator/Arguments.swift
```diff
@@ -202,7 +202,7 @@ func getArgRef (arg: JGodotArgument) -> String {
                 optstorage = ".content"
             } else {
                 needAddress = "&"
-                optstorage = ".handle"
+                optstorage = ".pNativeObject"
             }
```
migueldeicaza, line 206: Please do not change the `handle` name to the ugly looking pNativeObject, for one, I dislike the name and notation, but it is also a lot of churned code for this change.

Sources/SwiftGodot/Core/Wrapped.swift
```diff
@@ -95,9 +95,6 @@ open class Wrapped: Equatable, Identifiable, Hashable {
     public static var fcallbacks = OpaquePointer (UnsafeRawPointer (&Wrapped.frameworkTypeBindingCallback))
     public static var ucallbacks = OpaquePointer (UnsafeRawPointer (&Wrapped.userTypeBindingCallback))
     public static var deferred: Callable? = nil
-
-    /// Conformance to Identifiable by using the native handle to the object
-    public var id: Int { Int (bitPattern: handle) }
```
migueldeicaza, line 194: I do not understand this change, and the page for Swift's ObjectIdentifier is not great, but I suspect this might be fine nowadays, since we ensure Godot object identity via forcing everything via the lookupObject construct and do not do identity based on the handle anymore

Generator/Generator/ClassGen.swift
```diff
@@ -115,11 +115,11 @@ func generateVirtualProxy (_ p: Printer,
                 // This idiom guarantees that: if this is a known object, we surface this
                 // object, but if it is not known, then we create the instance
                 //
-                argPrep += "let resolved_\(i) = args [\(i)]!.load (as: UnsafeRawPointer?.self)\n"
+                argPrep += "let resolved_\(i) = args [\(i)]!.load (as: GDExtensionObjectPtr?.self)\n"
                 if isMethodArgumentOptional(className: cdef.name, method: methodName, arg: arg.name) {
```
migueldeicaza, line 119: What is the rationale for changing this pointer?

I suspect it is because this is what the handle is exported as, and it might make sense to use the same type Godot's extension is using, but the rationale for the UnsafeRawPointer was because the value was public, and this prevented an internal data type to be surfaced.

Generator/Generator/MethodGen.swift
```diff
@@ -754,7 +753,7 @@ func generateMethod(_ p: Printer, method: MethodDefinition, className: String, c
     
     p ("\(declarationTokens)(\(argumentsList))\(returnClause)") {
         if staticAttribute == nil {
-            p("if handle == nil { Wrapped.attemptToUseObjectFreedByGodot() }")
+            p("assertValidity()")
         }
```
migueldeicaza, line 756: The point of using this long-name was to spot it easily on crash reports that we are getting from users, it cuts down on the detective work, please do not rename it.

Sources/SwiftGodot/Core/Wrapped.swift
```diff
@@ -637,46 +647,65 @@ func unreferenceFunc(_ userData: UnsafeMutableRawPointer) {
...
+    let metatype = Unmanaged<AnyObject>.fromOpaque(userData).takeUnretainedValue()
+    
+    guard let metatype = metatype as? Object.Type else {
+        print("SwiftGodot.recreateFunc: The wrapped value does not contain a known type: \(metatype)")
         return nil
     }
-    bindingObject = godotObjecthandle
-    let o = type.init ()
-    bindingObject = nil
-    return UnsafeMutableRawPointer (mutating: o.handle)
```
migueldeicaza, line 682: What happened to bindingObject?   I wonder how the new code prevents the scenario that introduced this from commit 22a6fe50e

Sources/SwiftGodot/Extensions/PhysicsDirectSpaceState3D+IntersectRayResult.swift
```diff
@@ -30,7 +30,7 @@ extension PhysicsDirectSpaceState3D {
         /// The colliding object
         public let collider: T
         /// The colliding object's ID.
-        public let colliderId: Int
+        public let colliderId: UInt
         /// The The intersecting object's ``RID``.
```
migueldeicaza, line 34: What triggered this API change?

Sources/SwiftGodot/Core/Wrapped.swift
```diff
@@ -580,6 +550,22 @@ func existingSwiftObjectboundTo(_ pNativeObject: GDExtensionObjectPtr) -> Wrappe
     }
 }
 
+/// See `harmonizeReferenceCounting` for details
+enum RefTransferMode {
+    case retained
+    case unretained
+    case singleton
+
```
migueldeicaza, line 704: What is the reason to introduce this RefTransferMode?

Generator/Generator/BuiltinGen.swift
```diff
@@ -634,6 +634,7 @@ func generateBuiltinClasses (values: [JGodotBuiltinClass], outputDir: String?) a
             conformances.append ("ExpressibleByStringLiteral")
             conformances.append ("ExpressibleByStringInterpolation")
             conformances.append ("LosslessStringConvertible")
+            conformances.append("@unchecked Sendable")
```
migueldeicaza, line 637: Would love to understand the rationale for this change.

We generally have not adopted the new concurrency support in SwiftGodot, so I would love to know what problem this is solving.   That said, I suspect that StringNames could be bona-fide `Sendable`

Sources/SwiftGodotMacroLibrary/MacroGodot.swift
```diff
@@ -337,17 +337,27 @@ public struct GodotMacro: MemberMacro {
...
+                """
+                override \(raw: accessControlLevel) class var classInitializer: Void {
+                    let _ = super.classInitializer
+                    return _initializeClass
+                }
+                """,
+                
+                """
+                private static let actualClassName: StringName = "\(raw: processor.className)"                                
+                """,
```
migueldeicaza, line 353: This is going to be problematic for Xogot, as we would need this to not be cached, like it is above, and I am not sure how to propagate the information from the generator to the macro processor that has no idea of this.

migueldeicaza:
Copy/pasting from Slack:

> This PR is growing too large with too many things on it
> I enjoyed the init cleanup, but this now has blended lifecycle problems that are going to be very difficult to test and ensure they are correct
> It takes us a couple of weeks of iteration to find some of the ugly corner cases, so I am very reluctant to do those changes

> I really would like to separate these large renames from functionality

Sources/SwiftGodot/Core/Wrapped.swift
elijah-semyonov: It's a part of the `init` chain.

Godot comes with it this pointer.
Swift creates it within `convenience init()` and triggers the only chain.

Sources/SwiftGodot/Core/Wrapped.swift
elijah-semyonov: Because it allows to distinguish three different scenario of how a `Reference` transfer into the Swift world happens.
It's an iteration and I think it can be removed.
It helped me to see how that context propagates down the call chain, because previously Godot singleton wrappers were `init` and instantly `deinit` on every `shared` call.

Generator/Generator/ClassGen.swift
elijah-semyonov: I propose to use type aliases for mutable raw pointers because they are quite hard to follow.
Especially in a context, where a pointer to pointer is expected (ptrcall, for example).

Sources/SwiftGodot/Extensions/PhysicsDirectSpaceState3D+IntersectRayResult.swift
elijah-semyonov: `Identifiable` had `id`, that was removed. Now it uses Godot `Object.getInstanceId()`.

Generator/Generator/Arguments.swift
elijah-semyonov: Reverted

Sources/SwiftGodot/Core/Wrapped.swift
elijah-semyonov: Yes, it's exactly the rationale behind this removal. It did made sense before, because we could have multiple framework bindings from Swift side for a single Godot object, but it's not longer a case

Generator/Generator/BuiltinGen.swift
elijah-semyonov: Macro generated code disallowed using static `let`s for non-sendable types.
@preconcurrency import is a bit weird :)

Sources/SwiftGodotMacroLibrary/MacroGodot.swift
elijah-semyonov: In the same way as codegen - [SPM traits, when we support it.](https://forums.swift.org/t/pitch-package-traits/72191/14)

PR #700, opened 2025-04-28
Implement single init chain

elijah-semyonov:
### Single `required init` chain

There is only one `required init` chain now, that takes `InitContext`. In case users _do_ need a custom `init` they will have only:

```
    required init(_ context: InitContext) {
        super.init(context)
        
        addPoint (id: Self.A, position: Vector3 (x: 0, y: 0, z: 0))
        addPoint (id: Self.B, position: Vector3 (x: 1, y: 0, z: 0))
        addPoint (id: Self.C, position: Vector3 (x: 0, y: 1, z: 0))
        addPoint (id: Self.X, position: Vector3 (x: 0, y: 0, z: 1))
        connectPoints (id: Self.A, toId: Self.B)
        connectPoints (id: Self.A, toId: Self.C)
        connectPoints (id: Self.B, toId: Self.C)
        connectPoints (id: Self.X, toId: Self.A)
    }
 
```

They can't construct their own, it's required for `super.init`, so it's almost impossible to get wrong.

At the same time default initialiser `init()` is accessible for all `Object`-derived types, it will go through the same path but supply the newly constructed native object.

Initialiser `init(nativeHandle:)` is left, but now takes UnsafeMutableRawPointer

### Internal improvements

Global `bindingObject` is gone. Scope responsible for creation of `bindingObject` supplies it directly to the `required init` instead.

Opaque `Object *` is kept as `UnsafeMutableRawPointer` instead of `UnsafeRawPointer` to align with the GDExtension API. No more awkward `UnsafeMutableRawPointer(mutating: constRawPointer)`.

Minor renames.
`GodotNativeObjectPointer` typealias for `UnsafeMutableRawPointer`.

### DIFF
--- a/Sources/SwiftGodot/Core/Wrapped.swift
+++ b/Sources/SwiftGodot/Core/Wrapped.swift
@@ -27,6 +27,30 @@ func pd (_ str: String) {
 var xmap: [UnsafeRawPointer: String] = [:]
 #endif
 
+/// A specific place where the initialization happens
+/// affects how things unravel.
+enum InitOrigin {
+    /// Directly from Swift
+    /// For example: `let object = Object()`
+    ///
+    /// `RefCounted` performs an `initRef` in this patch.
+    case swift
+    
+    /// Constructed by Godot (from GDScript) in `createFunc`
+    /// Or returned back from Godot world via unwrapping a ``Variant`` or as return value from the function call
+    /// ``WrappedReference`` will `strongify` the wrapped Swift reference to avoid it being immediately destroyed
+    case godot
+}
+
+/// Opaque pointer representing Godot `Object *`
+public typealias GodotNativeObjectPointer = UnsafeMutableRawPointer
+
+/// Just pass it to `super.init`.
+public struct InitContext {
+    let handle: GodotNativeObjectPointer
+    let origin: InitOrigin
+}
+
 ///
 /// The base class for all class bindings in Godot, you should not have
 /// to instantiate or subclass this class directly - there are better options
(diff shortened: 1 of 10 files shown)


Sources/SwiftGodot/Core/Wrapped.swift
```diff
@@ -91,7 +115,10 @@ var xmap: [UnsafeRawPointer: String] = [:]
 /// call you back on will not be invoked.
 open class Wrapped: Equatable, Identifiable, Hashable {
     /// Points to the underlying object
-    public var handle: UnsafeRawPointer?
+    public var handle: GodotNativeObjectPointer?
+    
+    weak var wrapper: WrappedReference?
+
```
migueldeicaza, line 121: It looks like we store the reference here, but this stored referenced is never used.

Was this a left over?

elijah-semyonov: Oh, indeed, a nice catch. I've actually wanted to avoid a secondary lookup at the global table to strongly the reference.

elijah-semyonov: https://github.com/migueldeicaza/SwiftGodot/pull/700/commits/0e25e5abf0dda7aa6805f29a8ad3e0445edf1c91
That's what it is for :)

Sources/SwiftGodot/Core/Wrapped.swift
```diff
@@ -944,7 +979,7 @@ public func getActiveHandles() -> ([UnsafeRawPointer]) {
     return handles
 }
 
-public func clearHandles(_ handles: [UnsafeRawPointer]) {
+public func clearHandles(_ handles: [GodotNativeObjectPointer]) {
     for handle in handles {
```
migueldeicaza, line 983: This also needs the companion getActiveHandles to get the type change.

elijah-semyonov: Done. Is it fine that both of them do not acquire a lock?

migueldeicaza:
Reviewed and tested with SwiftGodotKit and Xogot, seems to work perfectly!

Issue #714, opened 2025-06-01
Custom Resources crash when editing in Godot, do not persist data changes


Naxum:
An incredibly simple custom Resource has several problems in recent SwiftGodot changes.

In Godot 4.4.1, attempting to create a file with this resource definition:
```
import SwiftGodot

@Godot
class Item: Resource {
    @Export
    var cost: Double = 100
}
```

1. Almost always crashes when attempting to edit an instance in Godot's UI
2. Does not persist value changes in Godot's UI
3. Does not persist value changes in an external text editor when opened in Godot

In this older commit, **Resources work correctly**:  9ef458124f6b5f51db45dfcf67fa041402705198
In the latest commit, the problem is present: 9085954b0c1284202f7b7d0c59cd4c2b92748fac

# Crash is as follows:

Crashed Thread:        0  Dispatch queue: com.apple.main-thread
Exception Type:        EXC_BAD_ACCESS (SIGABRT)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000069
Exception Codes:       0x0000000000000001, 0x0000000000000069
Termination Reason:    Namespace SIGNAL, Code 6 Abort trap: 6
Terminating Process:   Godot [38007]
VM Region Info: 0x69 is not in any region.  Bytes before following region: 4370382743
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
---> 
      __TEXT                      1047ec000-10bf7c000    [119.6M] r-x/r-x SM=COW  /Applications/Godot.app/Contents/MacOS/Godot
Application Specific Information:
abort() called
0   libsystem_kernel.dylib        	       0x182dbf388 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x182df88cc pthread_kill + 296
2   libsystem_c.dylib             	       0x182d01c60 abort + 124
3   Godot                         	       0x104fb6068 0x1047ec000 + 8167528
4   libsystem_platform.dylib      	       0x182e326a4 _sigtramp + 56
5   Godot                         	       0x1095b3cd8 Object::get(StringName const&, bool*) const + 164
6   Godot                         	       0x1095b3cd8 Object::get(StringName const&, bool*) const + 164
7   Godot                         	       0x10670ec10 EditorProperty::is_cache_valid() const + 104
8   Godot                         	       0x10673eff4 EditorInspector::_notification(int) + 2940
[...]

PR #715, opened 2025-06-01
Make `initRef` unconditional

elijah-semyonov:
This change fixes a [regression](https://github.com/migueldeicaza/SwiftGodot/issues/714) introduced by #700 due to a confusion of when `initRef` should be called in `Wrapped.init(nativePointer:)` and `Wrapped.init()` codepaths.


migueldeicaza:
Thank you Elijah!

PR #722, opened 2025-07-12
Remove unsafe flags save for debug

cosmicburp:
I get this error using SwiftGodot via CLI. Not sure if anyone else is seeing it but using unsafeFlags in non-debug is a FatalError. In the long term having no errors at all is great but I think there really a bunch of warnings. Will see what I can do in future PRs.


cosmicburp:
@migueldeicaza Okay makes sense. Looks like it was already commented out anyway. Should I just close this or should we just remove?

PR #723, opened 2025-07-12
Fix some warnings

cosmicburp:
I was looking through some of the warnings and some of them are easily fixable

### DIFF
--- a/Sources/SwiftGodot/Core/SignalWithArguments.swift
+++ b/Sources/SwiftGodot/Core/SignalWithArguments.swift
@@ -61,7 +61,7 @@ public struct SignalWithArguments<each T: _GodotBridgeable> {
     @discardableResult
     public func connect(flags: Object.ConnectFlags = [], _ callback: @escaping (_ t: repeat each T) -> Void) -> Callable {
         let callable = Callable(callback)
-        let error = target?.connect(signal: signalName, callable: callable, flags: UInt32(flags.rawValue))
+        _ = target?.connect(signal: signalName, callable: callable, flags: UInt32(flags.rawValue))
         return callable
     }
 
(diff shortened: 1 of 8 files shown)


Sources/SwiftGodot/Core/SignalWithArguments.swift
```diff
@@ -61,7 +61,7 @@ public struct SignalWithArguments<each T: _GodotBridgeable> {
     @discardableResult
     public func connect(flags: Object.ConnectFlags = [], _ callback: @escaping (_ t: repeat each T) -> Void) -> Callable {
         let callable = Callable(callback)
-        let error = target?.connect(signal: signalName, callable: callable, flags: UInt32(flags.rawValue))
+        _ = target?.connect(signal: signalName, callable: callable, flags: UInt32(flags.rawValue))
```
cosmicburp, line 64: I'd like to actually make this a throwing function but that would break everyone. Not sure what kind of errors we could expect from this anyhow.

Sources/SwiftGodot/Core/TypedArray.swift
```diff
@@ -186,7 +186,7 @@ public struct TypedArray<Element: _GodotContainerTypingParameter>: CustomDebugSt
         hasher.combine(hash())
     }
     
-    public static func ==(lhs: Self, rhs: Self) {
+    public static func == (lhs: Self, rhs: Self) -> Bool {
```
cosmicburp, line 189: This seems like a genuine error. As in the equatable wasn't implemented at all. Notably also now that I look at it the hash() implementation isn't mirroring the equatable implementation which might cause issues too. They should both use hash or both use array. This might not be a problem not I'd have to see how hash is generated

migueldeicaza:

I am a bit unsure about one of the changes on the private, but we will test it live :-)

PR #747, opened 2025-12-02
Add variant encoder/decoder with VariantConvertible conformance through VariantCodable

evanwang0:
## Summary
- Adds `VariantEncoder` and `VariantDecoder` — full implementations of Swift's `Encoder`/`Decoder` protocols that convert `Codable` types to and from Godot `Variant` values
- Structs encode to `VariantDictionary`, arrays encode to `VariantArray`, and primitives encode to their corresponding `Variant` types
- Supports the complete `Codable` contract: keyed/unkeyed/single-value containers, nested containers, optional values, enums with raw values, class inheritance via `superEncoder`/`superDecoder`, and all Swift numeric primitives
- Add `VariantCodable `marker protocol that refines both `Codable` and `VariantConvertible`, providing default implementations of all `VariantConvertible` requirements via the encoder/decoder

### Basic Struct

```swift
struct PlayerStats: VariantCodable, Equatable {
    let maxHealth: Int
    let attackSpeed: Double
    let isPlayerControlled: Bool
}

let stats = PlayerStats(maxHealth: 100, attackSpeed: 1.5, isPlayerControlled: true)
guard let variant = stats.toVariant() else { return }
// { "max_health": 100, "attack_speed": 1.5, "is_player_controlled": true }

let decoded = try PlayerStats.fromVariantOrThrow(variant)
```

### Enums

```swift
enum Direction: String, Codable {
    case north, south, east, west
}

struct WithEnum: VariantCodable {
    let name: String
    let direction: Direction
}

let compass = WithEnum(name: "compass", direction: .north)
guard let variant = compass.toVariant() else { return }
// { "name": "compass", "direction": "north" }
```

### Enums with Associated Values

```swift
// Labeled — uses the label names as keys
enum PowerUp: Codable {
    case healthBoost(maxHealthGain: Int)
    case damageBoost(attackDamageBonus: Int, durationSeconds: Double)
}

struct Player: VariantCodable {
    let playerName: String
    let activePowerUp: PowerUp
}

let player = Player(
    playerName: "Hero",
    activePowerUp: .damageBoost(attackDamageBonus: 25, durationSeconds: 10.0))
guard let variant = player.toVariant() else { return }
// {
//   "player_name": "Hero",
//   "active_power_up": {
//     "damage_boost": { "attack_damage_bonus": 25, "duration_seconds": 10.0 }
//   }
// }
```

Unlabeled associated values use `_0`, `_1`, etc.:

```swift
enum PowerUp: Codable {
    case healthBoost(Int)
}

// .healthBoost(50) -> { "health_boost": { "_0": 50 } }
```

### Key Strategies

```swift
// Default: .convertToSnakeCase for encoding, .convertFromSnakeCase for decoding
let variant = stats.toVariant()
// { "max_health": 100, "attack_speed": 1.5, "is_player_controlled": true }

// Use default keys (no conversion)
let encoder = VariantEncoder(keyEncodingStrategy: .useDefaultKeys)
try stats.encode(to: encoder)
// { "maxHealth": 100, "attackSpeed": 1.5, "isPlayerControlled": true }

let decoder = VariantDecoder(encoder.value, keyDecodingStrategy: .useDefaultKeys)
let decoded = try PlayerStats(from: decoder)
```

### Nested Types

```swift
struct Inventory: VariantCodable {
    let playerName: String
    let items: [PlayerStats]
}

let inventory = Inventory(
    playerName: "Hero",
    items: [
        PlayerStats(maxHealth: 100, attackSpeed: 1.5, isPlayerControlled: true),
        PlayerStats(maxHealth: 200, attackSpeed: 2.5, isPlayerControlled: false),
    ])
guard let variant = inventory.toVariant() else { return }
// {
//   "player_name": "Hero",
//   "items": [
//     { "max_health": 100, "attack_speed": 1.5, "is_player_controlled": true },
//     { "max_health": 200, "attack_speed": 2.5, "is_player_controlled": false }
//   ]
// }
```

### Inheritance with Super Encoder

```swift
class Animal: Codable {
    let name: String
    let legs: Int
}

class Dog: Animal, VariantCodable {
    let breed: String
}

let dog = Dog(name: "Rex", legs: 4, breed: "Labrador")
guard let variant = dog.toVariant() else { return }
// {
//   "breed": "Labrador",
//   "__super": { "name": "Rex", "legs": 4 }
// }
```

## Background
The Swift standard library defines a standardized approach to data encoding and decoding through the `Encodable` and `Decodable` protocols. Custom types adopt these protocols to allow `Encoder` and `Decoder` implementations to serialize them to and from external representations such as JSON or property lists. Conforming to `Codable` (which combines both protocols) is the idiomatic way to make Swift types serializable.

Similarly, SwiftGodot defines `VariantConvertible` — the core protocol for converting Swift types to and from Godot's `Variant`, the universal type used across the Godot engine API. Any type that needs to be converted to or from a `Variant` must conform to `VariantConvertible`. Types used in exported properties, function arguments, and signals need the deeper `_GodotBridgeable` conformance.

## Motivation
Currently, if you want to read custom Swift data structures in GDScript without using `@Godot` + `@Export`, you have to manually implement `VariantConvertible` — writing the conversion logic to pack each field into a `VariantDictionary` and unpack it back. This is tedious, error-prone, and duplicates what Swift's Codable system already does automatically.

## Open questions/discussions
- The original plan was to provide a default implementation of `VariantConvertible` for types that conform to both `Codable` and `VariantConvertible`, so that users could write struct `PlayerData: Codable, VariantConvertible {}` without implementing any of the conversion methods manually. However, this creates a compile-time witness table ambiguity: built-in types like `Int`, `String`, and `Bool` already have explicit `VariantConvertible` conformances and also conform to `Codable`, so the compiler cannot choose between the existing implementation and the new Codable-based implementation, resulting in build failures. I feel like it's sub-optimal to introduce a new `VariantCodable` type in order to provide the behavior for the standard `VariantConvertible` type but maybe my judgement is wrong. I hoped that Swift would allow me to specify an implementation priority but I could not find such a feature (notwithstanding that it maybe a footgun of a feature). One idea was to remove the existing conformances however I would assume there would be considerable performance impact. I'm open to suggestions here and I'm happy to attempt to follow through on them. 
- The other aspect I think worth mentioning is that VariantConvertible requires both encoding/decoding to/from Variant. I think it would be useful to emulate Swift's design and have it be a composition of VariantEncodable + VariantDecodable. That way consumers are not forced to implement bidirectional conversion if they do not need it. This is useful when performing a type erasure within Swift. I'm not sure of the implications that this would have to the rest of the API.

## Notes
- The default encoding/decoding key strategy is to convert camelCase to snakeCase and vice versa to match the current behavior for API conversion. This can be changed on request.
- When a class uses `superEncoder()/superDecoder()`, the superclass data is stored under the key `"__super"` in the dictionary. I selected this to avoid potential collisions but defer to the maintainers to decide on the canonical key. For reference, JSONEncoder uses `"super"`
- There is one instance (encoding super) where an encode needs to return an encoder to the caller. In effect, the system needs to link a reference to an object that will be filled out. This is accomplished using `writeBack` which is a callback which sets the value when the returned encoder de-initializes. This is based of the behavior in JSONEncoder

## Todo
- <strike>Add key encoding/decoding strategy (e.g. `snake_case`/`PascalCase`) similar to `JSONEncoder.KeyEncodingStrategy` to match GDScript/Godot conventions of snake case.</strike>


migueldeicaza:
This looks very interesting, would love to see this come to fruition.

evanwang0:
Coming back to the first open question, I noticed Swift 6.3 introduced function specialization:

<blockquote>
Function specialization: Provide pre-specialized implementations of a generic API for common concrete types using `@specialize`.
</blockquote>

This sounds like it maybe applicable to the problem at hand. I will take a look at this. My immediate concern is that it applies to generics which is not currently how `VariantConvertible`/`VariantCodable` are being implemented

PR #761, opened 2026-01-13
change names of variables in tests to '_' to remove warning errors

CCranney:
When building SwiftGodot, Swift advises changing the names of several variables in the `MemoryLeakTests.swift` tests to `_`, as they are unused. It appears that a few of them had commented-out code in their vicinity that would require those variables to be named, but are not used at this time. In case that commented-out code is useful in the future, line comments were included to advise changing the name back in such instances.

There remain deprecation warning errors in `Tests/SwiftGodotTestExtension/SignalTests.swift`  tests, advising the user to use the `@Signal` macro instead. It wasn't clear to me if these tests were designed to specifically ensure the old code is still working, or if the tests should be updated to reflect the macro usage.

### DIFF
--- a/Tests/SwiftGodotTestExtension/MemoryLeakTests.swift
+++ b/Tests/SwiftGodotTestExtension/MemoryLeakTests.swift
@@ -61,7 +61,7 @@ class GodotEncoder: Encoder {
     }
 
     func encode(key codingKey: [CodingKey], value: Variant) {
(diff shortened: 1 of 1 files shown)


CCranney:
Ah - I just saw #655 as a solution for the remaining Signal macro errors, which addresses my question about the remaining warning errors.

Issue #447, opened 2024-04-05
WebAssembly support


migueldeicaza:
We would like to make it so that developers that want to use SwiftGodot can target Wasm, a popular platform for Godot to be deployed to.

This issue is intended to keep track of the various findings, it will be edited as we find out more information, so this text will change.

Here is roughly where we stand right now:

1. Godot seems to use the Emscripten ABI
2. Swift/Wasm seems to use the WASI API, and these are different
3. We use a lot of Macros, and SwiftWasm does not currently support it, but there is a patch coming soon https://github.com/apple/swift-package-manager/pull/7118

While SwiftWASM only supports the executable target, it seems that in the Wasm world, even regular executables can be treated as libraries, so we would need to figure out a way of stitching those together.    One possible problem with this approach is that currently we might assume that the extension and Godot share the same memory address space so this could be a problem.

Things that we could also consider:

1. Compiling Godot as a library 
2. Have our build system link Godot with the SwiftGodot target, like SwiftGodotKit in a way.

# Discussion links

Initial discussion on using SwiftWasm for Godot:
https://discord.com/channels/780838335798706197/802944960881098752/1208500336671854662

Recent discussion:

https://discord.com/channels/780838335798706197/790246707648135229/1225627478471610398

# Resoures

Setting up Swift for WASM: https://book.swiftwasm.org/getting-started/setup.html

migueldeicaza:
There are a few updates shared on Slack.

1. The ABI between Godot and Swift was sorted out, the details of how are not clear to me yet

https://files.slack.com/files-pri/T05CA68KQMB-F07PZ7XAFQW/screenshot_2024-10-02_at_10.20.31___pm.png

https://files.slack.com/files-pri/T05CA68KQMB-F07PZ7XRF70/screenshot_2024-10-02_at_10.20.42___pm.png

migueldeicaza:
This command line will build a file that can be linked with Godot Wasm:

```
swiftc -Osize -Xcc -fdeclspec -target wasm32-unknown-none-wasm -enable-experimental-feature Extern -enable-experimental-feature Embedded -wmo -c -Xlinker --no-entry -Xclang-linker -nostdlib -parse-as-library HelloWorld.swift -o HelloWorld.wasm
```

From @jordo 

[...]

PR #762, opened 2026-01-13
Add WASM compilation support

Maxim-Lanskoy:
This PR enables SwiftGodot to compile for WebAssembly target (wasm32-unknown-wasip1)

## Changes

1. **Add WASILibc import** - C library detection for WASI platform
2. **Fix OptionSet rawValue type** - `Int` → `Int64` to prevent overflow on 32-bit WASM
3. **Fix debug sentinel values** - `0xdeaddead` → `0` (overflows 32-bit Int)
4. **Fix PropertyUsageFlags init** - Match Int64 rawValue type

## Why

WASM32 has 32-bit `Int`, causing overflow when:
- Large enum flag values are used in OptionSet
- Debug sentinel `0xdeaddead` (3,735,936,685) is assigned to pointer

## Notes

- These changes don't affect existing platforms (macOS, iOS, Linux, Windows)
- Full runtime support requires toolchain-level work (Swift WASM SDK uses WASI ABI, Godot uses Emscripten ABI) - #447 for details

<details>
<summary><b>How to build SwiftGodot for WASM</b></summary>

### Setup

```bash
# Swift 6.2.3 via swiftly
curl -sL https://swiftlang.github.io/swiftly/swiftly-install.sh | bash
swiftly install 6.2.3
swift sdk install swift-6.2.3-RELEASE_wasm

# Emscripten 3.1.74 (Godot's version)
git clone https://github.com/emscripten-core/emsdk.git ~/emsdk
cd ~/emsdk && ./emsdk install 3.1.74 && ./emsdk activate 3.1.74
```

### Build

```bash
# Compile Swift to WASM
swift build --swift-sdk swift-6.2.3-RELEASE_wasm --target SwiftGodot

# Link with Emscripten as side module (adds dylink.0 section)
emcc -sSIDE_MODULE=2 -sEXPORTED_FUNCTIONS=_swift_entry_point \
    -sALLOW_MEMORY_GROWTH=1 -sERROR_ON_UNDEFINED_SYMBOLS=0 -O2 \
    .build/wasm32-unknown-wasip1/debug/*.build/*.o \
    ~/Library/org.swift.swiftpm/swift-sdks/swift-6.2.3-RELEASE_wasm.artifactbundle/swift-6.2.3-RELEASE_wasm/wasm32-unknown-wasip1/swift.xctoolchain/usr/lib/swift_static/wasi/libswiftCore.a \
    -o SwiftLogic.wasm
```

This produces a WASM binary with `dylink.0` section and `swift_entry_point` exported.

### Current limitation

The binary compiles but doesn't run in Godot's web export due to ABI mismatch between Swift WASM SDK (WASI) and Emscripten.

</details>

### DIFF
--- a/Generator/Generator/Enums.swift
+++ b/Generator/Generator/Enums.swift
@@ -67,8 +67,8 @@ func generateEnums (_ p: Printer, cdef: JClassInfo?, values: [JGodotGlobalEnumEl
             let optionTypeName = getGodotType (SimpleType (type: enumDef.name))
             var optionNames: [String] = []
             p ("public struct \(optionTypeName): OptionSet, CustomDebugStringConvertible") {
-                p ("public let rawValue: Int")
-                p ("public init (rawValue: Int)") {
+                p ("public let rawValue: Int64")
+                p ("public init (rawValue: Int64)") {
                     p ("self.rawValue = rawValue")
                 }
                 for enumVal in enumDef.values {
--- a/Generator/Generator/Printer.swift
+++ b/Generator/Generator/Printer.swift
@@ -47,6 +47,8 @@ class Printer {
         import Glibc
         #elseif canImport(Musl)
         import Musl
+        #elseif canImport(WASILibc)
+        import WASILibc
         #else
         #error("Unable to identify your C library.")
         #endif
--- a/Sources/SwiftGodotRuntime/Core/Wrapped.swift
+++ b/Sources/SwiftGodotRuntime/Core/Wrapped.swift
@@ -879,7 +879,7 @@ func validatePropertyFunc(ptr: UnsafeMutableRawPointer?, _info: UnsafeMutablePoi
     let className = StringName(fromPtr: classNamePtr)
     let hint = PropertyHint(rawValue: Int64(info.hint)) ?? .none
     let hintStr = GString(content: infoHintPtr.load(as: Int64.self))
-    let usage = PropertyUsageFlags(rawValue: Int(info.usage))
+    let usage = PropertyUsageFlags(rawValue: Int64(info.usage))
 
     var pinfo = PropInfo(propertyType: ptype, propertyName: pname, className: className, hint: hint, hintStr: hintStr, usage: usage)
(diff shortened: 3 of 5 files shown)


Maxim-Lanskoy:
Good question! I've looked more carefully at this and here are key findings:

The generated regular enums already use `Int64` as rawValue (Enums.swift line 107), so OptionSets using `Int` is actually an inconsistency in the current code. The `writeResult` infrastructure in Arguments.swift already handles both `Int` and `Int64` rawValues, and the values are passed to GDExtension API as `UInt32` regardless.

That said, you're right that changing OptionSet rawValue from `Int` to `Int64` could break user code that explicitly accesses `.rawValue` and expects `Int` — for example `let x: Int = flags.rawValue` or generic constraints like `where T.RawValue == Int`.

In practice, no current Godot enum values actually overflow 32-bit `Int` (the largest is `KeyModifierMask` at `0x7f000000`), so this change is precautionary rather than fixing a current bug.

I'm happy to drop this change from the PR if it's too disruptive. The other three fixes (WASILibc import, `0xdeaddead` sentinel overflow, and the `Int()` → `Int64()` match in Wrapped.swift) are the ones that actually prevent compilation on WASM. The OptionSet change could be revisited separately if/when it becomes necessary.

On a related note — there's now a pitch for Emscripten target support in Swift (https://forums.swift.org/t/pitch-emscripten-target-support-for-swift/85310) on the Swift forums, which explicitly cites SwiftGodot as a motivating use case. If that lands, it would solve the fundamental ABI mismatch (WASI vs Emscripten) that's the real blocker for WASM support. Even with Emscripten target though, `Int` would still be 32-bit on wasm32, so this consistency question remains — just not urgently.

migueldeicaza:
Thanks for all the detailed information, I would love if you could split the patch in two, the parts that are not controversial or would cause any issues we could get in right away, the other part we can discuss separately, and prepare the community for the change before we land.

I was very excited about the Swift announcement recently, hope this can help us get closer to getting Swift on Godot on the Web.

PR #763, opened 2026-01-14
ExtensionApi Refactor

CCranney:
## Goal
Refactoring `ExtensionApi` to use an unmodified version of [quicktype](https://app.quicktype.io/)'s output (`Sources/ExtensionApi/ApiJsonModel.swift`).

## Problem Overview
I've been working with SwiftGodot for a few weeks, but needed to adapt it to a different version of Godot for my use cases. In general, I found there were 4 main files that may need to be updated with a Godot update:

1. `Sources/ExtensionApi/extension_api.json`, from the godot engine
2. `Sources/ExtensionApi/ApiJsonModel.swift`, output from quicktype based on `extension_api.json`
3. `Plugins/CodeGeneratorPlugin/plugin.swift`, as newly generated classes must be explicitly included for the plugin
4. `Sources/GDExtension/gdextension_interface.h`, from the godot engine

The `ApiJsonModel.swift` took a minute to parse out, as it had been heavily modified from the original quicktype output to accommodate various workflows in SwiftGodot. I needed to separate the manual changes from the versioning changes by hand. I made my own manual updates to make it work, but wondered if it would be possible to streamline the process by refactoring the manual edits out into the neighboring `ApiJsonModel+Extra.swift` file. I dug into what that process may entail, and this PR is a proposal for that refactoring.

## Possible Alternatives
This is the "nuclear" option - a refactor that enables SwiftGodot to use the quicktype output without any modifications. Thus I understand if putting up with manual modifications of `ApiJsonModel.swift` is the preferred approach, as the refactoring can cause unexpected regressions or introduce other unwanted behavior. The two main alternatives I thought of are: 

1. Organize the modifications in such a way that adapting a new file on the fly will be relatively straightforward while leaving the logic untouched. This would be cosmetic update, making adaptations from quicktype more readable should someone need to adapt to a new godot version.
2. Put some debug instructions in the documentation. I found the fastest way to add my own manual edits was by running `swift build` with the old `ApiJsonModel.swift` file but new `extension_api.json` file and manually fixing the errors. It could be that instruction would be sufficient.

## Edit Summary

This build passed all tests by running:

```bash
swift package reset
swift build
swift run SwiftGodotTestRunner
swift test
```

1.  Refactor `JGodotBuiltinClassEnum` and `JGodotGlobalEnumElement` to use a shared protocol, `JGodotEnum` (d5594a4). This is given special attention because my approach utilizes the `associatedtype` keyword when referencing `JGodotEnum` variables, which may not be desirable behavior - warning errors were addressed in (411c9b4).
2. Adjust variable ordering and struct naming - a cosmetic change that removes `git diff` highlights (8db5685).
3. Add protocols and `typealias`'s to support differing naming convention (8db5685, 2c8f271). These can be reverted by adjusting the SwiftGodot code directly, but these were included to preserve possibly desired naming conventions. One of these does make use of `associatedtype`, which may not be desirable behavior.
4. Mark `CodingKeys` on all relevant classes to `public` (a2e7aea). End users shouldn't have need for this, it is purely to match `quicktype`'s output.
5. extended JSONNull to use a `hash` function (3e800be). I believe the quicktype output use of `hashValue` is deprecated, and this is the preferred method. It was manually inserted before, now it's addressed via extension.
6. Similar JGodot class functionality was shared via protocol rather than replacement (cb7b00d).
7. `enum`s were refactored to be manually checked (0bb06aa, following commits). Originally, quicktype provided several `enum`s. However, they were often heavily edited or removed entirely. This refactor included ticking the quicktype option of providing enums to "OFF", and enforced enum constraints in SwiftGodot code directly. This is perhaps the most extensive refactor behavior - tests were written to confirm functionality long-term (`Tests/SwiftGodotUniversalTests/EnumValidationTests.swift`).
8. Optional vs non-optional modified behavior was addressed via extensions and variable renaming (3a135c0).
9. The `SimpleType` struct defined in `Generator/Generator/TypeHelpers.swift` was edited to utilize this enum enforcement (d6bb9ef).

Note: After commit 3a135c0, I find that the following errors sometimes appear during the build. It's random, from what I can tell - most of the time these errors do not appear. I suspect the program tries to access files that have not been generated yet, explaining the randomness element. They do not seem to impact the build itself, and it has yet to cause problems, so have not directly addressed them, but thought I would point it out here.

```
 Internal Error: dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "Corrupted JSON", underlyingError: Optional(unexpected end of file)))
```

## Final Summary
I was able to copy-and-paste the quicktype output into ApiJsonModel.swift directly, and it built and passed the tests without error. However, these changes were not for technical improvement, they just adapted functionality or wrote code to match quicktype outputs directly. As such, I would completely understand if this whole concept should be scrapped. If the idea is sound, but the "nuclear" option specifically is not desired, this PR would still serve as a good reference for some of the proposed alternatives. Regardless, thank you for developing this project and making it open source!

## Settings used for generating [quicktype](https://app.quicktype.io/) output

1. Change `Name` field to `ExtensionApi`.
11. `Prefix for type names` -> `JGodot`
12. `Access level` -> `Public`
13. `Detect enums` -> OFF (see Edit Summary 7 for details - this differs from prior settings)

### DIFF
--- a/.github/workflows/generator.yml
+++ b/.github/workflows/generator.yml
@@ -30,12 +30,13 @@ jobs:
         filters: |
           generator:
             - 'Generator/**'
+            - '.github/workflows/generator.yml'
 
   # JOB to run the diff - only runs if changes were detected in Generator/
   diff-generator:
     needs: changes
     if: ${{ needs.changes.outputs.generator == 'true' }}
-    runs-on: macos-14
+    runs-on: macos-26
 
     steps:
     - name: Checkout
@@ -48,10 +49,11 @@ jobs:
         filters: |
           generator:
             - 'Generator/**'
+            - '.github/workflows/generator.yml'
     - name: Set up Xcode
       uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: "16.2"
+        xcode-version: "26.2"
     - name: Build This Branch
       run: swift build --target SwiftGodot --scratch-path .build/New
     - name: Checkout Target Branch


CCranney:
It looks like a check failed because the `generator.yml` file is dependent on old macOS and Xcode versions (14 and 16.2, respectively) - I created #765 to address the thrown error.

migueldeicaza:
My fear is the following, this requires me to do a bunch of testing right now, while I still have a few branches across multiple repositories to merge so I am not particularly thrilled for this major change.

Additionally, if the JSON file changes more substantially there are plenty of other problems to worry about, beyond just parsing the file.   I would be happy to integrate some small changes in the form of optional components to the json file, but a large refactor at this point will just have to wait.   But I would like to accommodate some of your workflow needs.

CCranney:
For any interested, I created a single-file update that manually marked each manual edit of `ApiJsonModel.swift`. I've opted not to make a PR for it - it's a niche variation of this PR - but if you're trying to parse out manual edits from version updates, this file could be helpful. It can be dropped in place of the current `ApiJsonModel.swift` file as-is. 

[ApiJsonModel.swift.tar.gz](https://github.com/user-attachments/files/24691114/ApiJsonModel.swift.tar.gz)

General outline of tasks to make this work:
1. Add this file to a branch of SwiftGodot in place of the original `ApiJsonModel.swift` file. Commit the change.
2. Create a quicktype output of the new godot version's `extension_api.json` file. Use `git diff` to see if any new changes are NOT marked with a "modification" note, indicating a version-specific change, like so:

Screenshot 2026-01-17 at 9 40 56 AM

For direct comparison, these are the settings I used for generating [quicktype](https://app.quicktype.io/) output
1. Change `Name` field to `ExtensionApi`.
4. `Prefix for type names` -> `_JGodot` (instead of `JGodot` - to delineate changes in the file from changes to SwiftGodot code at large, I prefixed it with an underscore `_`)
5. `Access level` -> `Public`

PR #770, opened 2026-01-28
Prevent Godot API calls during shutdown cleanup

elijah-semyonov:
Workaround Godot crash with code 6 during shut down


elijah-semyonov:
CI did run red (couldn't reproduce it locally). Assuming the hacky nature of this fix, I just decided to close it altogether. And w8 to see what the other fix you mentioned brings.

migueldeicaza:
Let me share the patch, it is much smaller than I had imagined:

https://gist.github.com/migueldeicaza/257c4644aca528d847b4c55b534dc79f

PR #779, opened 2026-04-02
Fix RefCounted leak: unreference Ref<> extra ref on new wrapper creation

### LINKED ISSUE #778: RefCounted objects returned via Ref<> create STRONG wrappers that never release
## Summary

When Godot methods return RefCounted objects via C++ `Ref<>` wrappers (e.g. `duplicate()`, `Image.create()`, `ResourceLoader.load()`, property getters), the returned object's reference count is inflated by 1. This causes `bindSwiftObject()` to create a STRONG `WrappedReference` that never transitions to WEAK, preventing the object from being freed during engine shutdown.

## Reproduction

```swift
let mat = StandardMaterial3D()
mat.albedoColor = Color(r: 1, g: 0, b: 0, a: 1)

let dup = mat.duplicate() as! StandardMaterial3D
print(dup.getReferenceCount())  // Prints 2 (should be 1)

// Set on a mesh and render...
meshInstance.materialOverride = dup

// On shutdown: mesh is freed → material refCount drops from 3 to 2
// Reference callback sees rc=2 (post-decrement), rc > 1 → weakify() NOT called
// Wrapper stays STRONG → material never freed → shader variant leaks
// DEV_ASSERT: "shaders of type SceneForwardClusteredShaderRD were never freed"
```

Compare with a directly created object:
```swift
let mat = StandardMaterial3D()
print(mat.getReferenceCount())  // Prints 1 (correct)
// On shutdown: weakify() IS called → wrapper freed → clean exit
```

## Root Cause

In `getOrInitSwiftObject()` (Wrapped.swift), there are two paths:

**Path A — existing wrapper** (line 726-728):
```swift
handleRef(staticType: T.self, object: object, ownsRef: ownsRef, unref: true)  // ← correct
```

**Path B — new wrapper** (lines 743, 752, 767):
```swift
handleRef(staticType: T.self, object: created, ownsRef: ownsRef, unref: false)  // ← bug
```

When `ownsRef: true` (Godot returned a `Ref<>`), Path A correctly calls `unreference()` to drop the extra ref. Path B does not, leaving the refCount inflated.

This causes `bindSwiftObject()` to see `refCount > 1` → STRONG wrapper. The Godot reference callback's `weakify()` condition (`rc == 1` post-decrement) is never met because the refCount is always 1 higher than expected.

## Impact

- **519 Godot API methods** return RefCounted types via `Ref<>` and are affected
- All `duplicate()`, `create*()`, `load*()`, `generate*()` methods, plus most property getters
- Objects leak at shutdown, causing `DEV_ASSERT` crashes in dev builds
- In release builds, the leak is silent but consumes memory

## Fix

Change `unref: false` to `unref: ownsRef` on the three new-wrapper paths:

```diff
- handleRef(staticType: T.self, object: created, ownsRef: ownsRef, unref: false)
+ handleRef(staticType: T.self, object: created, ownsRef: ownsRef, unref: ownsRef)
```

This makes new wrappers behave identically to existing wrappers when receiving `Ref<>` returns.

## Tested

- 19 API methods across 5 categories (duplicate, create, generate, getter, tween) verified with refCount assertions
- Clean shutdown (exit code 0) confirmed with `dev_build=yes` Godot builds
- No regressions observed in normal rendering/interaction 

## Related

- #36 — Original identification of the Ref<> ownership transfer issue
- #513 — Two sets of leaks (similar root cause in `asObject`)
- #663 — New memory management system (introduced `WrappedReference` but left this path unfixed)
- #672 — SwiftGodot Variant Leaks
- #683 — Bring back the clean teardown

xarple:
## Problem

When Godot methods return RefCounted objects via C++ `Ref<>` (519 methods including e.g. `StandardMaterial3D.duplicate()`, `Image.create()`, `ResourceLoader.load()`, property getters), the `Ref<>` wrapper adds an extra `reference()` call. For **existing** Swift wrappers, `handleRef` correctly drops this extra ref (`unref: true`). For **newly created** wrappers, `handleRef` was called with `unref: false`, leaving the refCount inflated by 1.

This caused `bindSwiftObject()` to create a STRONG `WrappedReference` (because `refCount > 1`). During shutdown, when Godot releases its reference, the reference callback's `weakify()` condition (`rc == 1`) was never met because the refCount was always 1 higher than expected. The wrapper stayed STRONG forever, leaking the Godot object and its resources.

## The Fix

```diff
// Wrapped.swift — getOrInitSwiftObject(), 3 new-wrapper paths:
- handleRef(staticType: T.self, object: created, ownsRef: ownsRef, unref: false)
+ handleRef(staticType: T.self, object: created, ownsRef: ownsRef, unref: ownsRef)
```

When `ownsRef: true` (returned via `Ref<>`), `handleRef` now calls `unreference()` to normalize the refCount — matching the existing-wrapper path's behavior. When `ownsRef: false`, behavior is unchanged.

Fixes #778

### DIFF
--- a/Sources/SwiftGodotRuntime/Core/Wrapped.swift
+++ b/Sources/SwiftGodotRuntime/Core/Wrapped.swift
@@ -740,7 +740,7 @@ public func getOrInitSwiftObject<T: Object>(nativeHandle: GodotNativeObjectPoint
     }
     if let userType = userTypes[className] {
         let created = userType.init(InitContext(handle: nativeHandle, origin: .godot))
-        handleRef(staticType: T.self, object: created, ownsRef: ownsRef, unref: false)
+        handleRef(staticType: T.self, object: created, ownsRef: ownsRef, unref: ownsRef)
         if let result = created as? T {
             return result
         } else {
@@ -749,7 +749,7 @@ public func getOrInitSwiftObject<T: Object>(nativeHandle: GodotNativeObjectPoint
     }
     if let ctor = lookupGodotType(named: className) as? T.Type {
         let result = ctor.init(InitContext(handle: nativeHandle, origin: .godot))
-        handleRef(staticType: T.self, object: result, ownsRef: ownsRef, unref: false)
+        handleRef(staticType: T.self, object: result, ownsRef: ownsRef, unref: ownsRef)
         return result
     }
 
@@ -764,7 +764,7 @@ public func getOrInitSwiftObject<T: Object>(nativeHandle: GodotNativeObjectPoint
     // This comment here for future generations that end up in this line
     // the real error is likely a registration issue.
     let result = T(InitContext(handle: nativeHandle, origin: .godot))
-    handleRef(staticType: T.self, object: result, ownsRef: ownsRef, unref: false)
+    handleRef(staticType: T.self, object: result, ownsRef: ownsRef, unref: ownsRef)
     return result
 }
 


migueldeicaza:
Could you please try the latest version out of main, I just merged all of the Xogot changes (where had done work on this area) and it should fix your scenario.

PR #781, opened 2026-04-12
Swift 6.3 compatibility: fix ambiguous operator overloads and bump to…

evanwang0:
…ols version

It seems 6.3 has become stricter about literal type conversion.
From: https://developer.apple.com/documentation/swift/expressiblebyintegerliteral

> The standard library integer and floating-point types, such as Int and Double, conform to the ExpressibleByIntegerLiteral protocol. You can initialize a variable or constant of any of these types by assigning an integer literal.

Adds @_disfavoredOverload to generated Double-parameter operator overloads when an Int-parameter overload also exists for the same operator. This resolves "ambiguous use of operator" errors in Swift 6.3 where numeric literals could match either overload. Also bump swift-tools-version to 6.3.

The alternative here is to force consumers to be explicit about the literal type however this would be a breaking change.

Also bump swift-tools-version to 6.3 and regenerate macro test expected outputs to match Swift 6.3 formatting.

### DIFF
--- a/Generator/Generator/BuiltinGen.swift
+++ b/Generator/Generator/BuiltinGen.swift
@@ -384,7 +384,12 @@ func generateBuiltinOperators (_ p: Printer,
             if let desc = op.description, desc != "" {
                 doc (p, bc, desc)
             }
-            
+
+            let disfavored = right == "float" && operators.contains(where: { $0.name == op.name && $0.rightType == "int" })
+            if disfavored {
+                p ("@_disfavoredOverload")
+            }
+
             p ("public static func \(swiftOperator)(lhs: \(lhsTypeName), rhs: \(rhsTypeName)) -> \(retType) "){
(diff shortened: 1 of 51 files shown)


migueldeicaza:
Thank you for sending this along, it generally looks fine, let me test it locally

evanwang0:
Ofc, let me know of any adjustment requests or validation requirements

migueldeicaza:
Apologies, this has taken me too long, but it seems like in the meantime the issues vanished?   I can not replicate the build failure.

evanwang0:
Yes, it looks like the changes you've made here https://github.com/migueldeicaza/SwiftGodot/commit/0d2e8a3d7def5bfbf736d929c39ede0e0697290c have resolved the issue. Given you've changed the tests, I believe it would affect consumers but it does address the build issues. I will close this PR.

PR #792, opened 2026-06-18
Auto-convert Swift names to Godot convention via SPM trait

elijah-semyonov:
Introduce the `automatic_godot_naming_convention` package trait (on by default) so every Swift identifier SwiftGodot registers with Godot follows the engine's convention: snake_case for methods/properties/signals/arguments, UPPER_SNAKE_CASE for enum constants, PascalCase classes unchanged.

The macros unconditionally emit calls to two non-@inlinable runtime functions (`_convertMemberNameToMatchGodotConvention`, `_convertEnumCaseNameToMatchGodotConvention` in the new GodotNaming.swift). Only those functions carry the `#if SWIFTGODOT_GODOT_NAMING` branch, and the trait sets that define on the runtime, so one switch flips behavior for every consumer. The converter is a Foundation-free scanner with byte-parity to the legacy camelToSnake regex (incl. 2D/3D fixups).

Conversion is applied at every registration site: @Callable methods + args, @Export property + get_/set_ accessors, @Signal name + args, @Rpc config name, #signal/SignalAttachment, nested-enum cases (_registerEnumIfPossible), the @Export enum PROPERTY_HINT_ENUM string (enumCasesHintStr), and manual ClassInfo registration. `@Callable(autoSnakeCase:)` is deprecated and ignored.

Fix two latent @Rpc bugs surfaced by the first compiled @Rpc usage: declare `_before_ready` in @Godot's member names, and build the rpcConfig dictionary imperatively (VariantDictionary isn't ExpressibleByDictionaryLiteral).

Tests: converter unit tests with legacy-regex parity; regenerated macro expansion fixtures; new NamingConventionTests asserting registered names against live Godot ClassDB. Migrate the embedded suite's string-based call sites and enum-case expectations to the Godot convention.

### DIFF
--- a/.github/workflows/docc.yml
+++ b/.github/workflows/docc.yml
@@ -22,10 +22,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      # Xcode 26.5 bundles Swift 6.3.2, which satisfies the package's
+      # swift-tools-version 6.3 (Xcode 26.2/26.3 only ship Swift 6.2.x).
       - name: Set up Xcode
         uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: "26.0"
+          xcode-version: "26.5"
 
       - name: Configure Pages
         uses: actions/configure-pages@v5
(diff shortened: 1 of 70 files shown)


elijah-semyonov:
I'll close this and reopen with static descriptors I mentioned in Discord.

PR #797, opened 2026-06-22
Add support for SwiftGodot classes to supply custom properties

ReddishBearGames:
There are several methods on Object that aren't virual, and so SwiftGodot creates them as `final` when generating the Swift interface. These methods are `getPropertyList()`, `propertyCanRevert()`, `propertyGetRevert()`, `get()` and `set()`

Overridding all of these is neccesary however if one wants to support custom properties in some fashion, and is a capability that other GDExtensions have (EG: C++).

The GDExtension API has function hooks for all of these methods that SwiftGodot does not supply today.

This PR adds hooks for all of these functions, and adds underbar versions of all the methods to the base Object. (`_getPropertyList()`, `_propertyCanRevert()`, `_propertyGetRevert()`, `_get()` and `_set()`)
These methods are overridable by subclasses to allow them to provide the neccesary functionality, but the default implementations don't alter the overall behavior for cases where no override is needed.

### DIFF
--- a/Sources/SwiftGodotRuntime/Core/ClassServices.swift
+++ b/Sources/SwiftGodotRuntime/Core/ClassServices.swift
@@ -296,24 +296,24 @@ public struct PropInfo: CustomDebugStringConvertible {
     }
 }
 
-class PropertyInfoStorage {
-	let propertyName: StringName
-	let className: StringName
-	let hintStr: GString
-	
-	var propertyNameContentPtr: UnsafeMutableRawPointer?
-	var classNameContentPtr: UnsafeMutableRawPointer?
-	var hintStrContentPtr: UnsafeMutableRawPointer?
-	
-	init(propertyName: StringName, className: StringName, hintStr: GString) {
-		self.propertyName = propertyName
-		self.className = className
-		self.hintStr = hintStr
-		
-		propertyNameContentPtr = withUnsafePointer(to: &propertyName.content) { UnsafeMutableRawPointer(mutating: $0) }
-		classNameContentPtr = withUnsafePointer(to: &className.content) { UnsafeMutableRawPointer(mutating: $0) }
-		hintStrContentPtr = withUnsafePointer(to: &hintStr.content) { UnsafeMutableRawPointer(mutating: $0) }
-	}
+public class PropertyInfoStorage {
+    let propertyName: StringName
+    let className: StringName
+    let hintStr: GString
+    
+    var propertyNameContentPtr: UnsafeMutableRawPointer?
+    var classNameContentPtr: UnsafeMutableRawPointer?
+    var hintStrContentPtr: UnsafeMutableRawPointer?
+    
+    init(propertyName: StringName, className: StringName, hintStr: GString) {
+        self.propertyName = propertyName
+        self.className = className
+        self.hintStr = hintStr
+        
+        propertyNameContentPtr = withUnsafeMutablePointer(to: &propertyName.content) { UnsafeMutableRawPointer($0) }
+        classNameContentPtr = withUnsafeMutablePointer(to: &className.content) { UnsafeMutableRawPointer($0) }
+        hintStrContentPtr = withUnsafeMutablePointer(to: &hintStr.content) { UnsafeMutableRawPointer($0) }
+    }
 }
 
 func bind_call (_ udata: UnsafeMutableRawPointer?,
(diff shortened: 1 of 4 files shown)


Sources/SwiftGodotRuntime/Core/ClassServices.swift
```diff
@@ -281,6 +296,26 @@ public struct PropInfo: CustomDebugStringConvertible {
...
+	var propertyNameContentPtr: UnsafeMutableRawPointer?
+	var classNameContentPtr: UnsafeMutableRawPointer?
+	var hintStrContentPtr: UnsafeMutableRawPointer?
+	
+	init(propertyName: StringName, className: StringName, hintStr: GString) {
+		self.propertyName = propertyName
+		self.className = className
+		self.hintStr = hintStr
+		
+		propertyNameContentPtr = withUnsafePointer(to: &propertyName.content) { UnsafeMutableRawPointer(mutating: $0) }
```
elijah-semyonov: `withUnsafeMutablePointer(to: &propertyName.content) { $0 }`
is better (same for 2 statements below)

Sources/SwiftGodotRuntime/Core/Wrapped.swift
```diff
@@ -306,6 +308,57 @@ open class Wrapped: Equatable, Identifiable, Hashable {
...
     }
+    
+    /// Override this if you want to have a property with special handling, perhaps the actual value gets stored elsewhere.
+    /// - Parameters:
+    ///     - property: The name of the property to set
+    ///     - value: A Variant representing the value to be stored.
+    /// - Returns:
+    /// Return `true` if you've handled the property and stored the value supplied - or discarded it but don't wish for the standard property storage to apply.
+    /// Return `false` if you don't have any special handling for the named property. Godot will store it in the usual way.
+    open func _set(property: StringName, value: consuming Variant?) -> Bool {
```
elijah-semyonov, line 324: `consuming` is no-op here (artifact from `FastVariant` and can be safely removed

Sources/SwiftGodotRuntime/Core/Wrapped.swift
```diff
@@ -306,6 +308,57 @@ open class Wrapped: Equatable, Identifiable, Hashable {
...
+        return nil
+    }
+    
+    /// Override if you want to allow reverting for the named property. You should also override `_propertyGetRevert()` Note that you do NOT have to return
+    /// `true` for any properties that you are not applying special handling to. Standard properties and any `@Export` properties will be handled appropriately even
+    /// if you return `false` here.
+    ///  - Returns:
+    ///  `true` if the named property supports reverting - this will almost always result in a subsequent call to `_propertyGetRevert()`
+    ///  `false` if the property does not support reverting.
+    open func _propertyCanRevert(_ property: StringName, canRevert: inout Bool) -> Bool {
```
elijah-semyonov: It's a tristate situation similar to `_get(property:)`. I'd introduce enum instead.

Sources/SwiftGodotRuntime/Core/Wrapped.swift
```diff
@@ -1195,6 +1267,135 @@ func validatePropertyFunc(ptr: UnsafeMutableRawPointer?, _info: UnsafeMutablePoi
...
+                return 1
+            }
+    }
+    return 0
+}
+
+// These exist to hold references to property list objects that are passed back to Godot,
+// but we retain ownership until `propertyListFreeFunc` is invoked, or the class is deinit'ed.
+var propertyListStorageCache: [PropertyInfoStorage]? = nil
+var propertyListPropInfoCache: [GDExtensionPropertyInfo]? = nil
```
elijah-semyonov: Shouldn't these be per object-instance caches?

elijah-semyonov:
Naming regarding `property` labeling is inconsistent, it's `property` in some places and `_ property` in others, let's choose one?

Sources/SwiftGodotRuntime/Core/Wrapped.swift
ReddishBearGames: Ooof, yes, I was under the impression that Wrapped was somehow associated to an instance but I'm not sure where I got that idea looking back. Nice catch!

Sources/SwiftGodotRuntime/Core/Wrapped.swift
ReddishBearGames: Agreed, done in next commit.

Sources/SwiftGodotRuntime/Core/Wrapped.swift
ReddishBearGames: Actually, upon further testing, removing this causes a 100% reproducible crash whenever this method is used. I can't actually explain *why* at the moment though.

PR #798, opened 2026-06-24
Add an "explicitName" argument to @Callable and @Export

ReddishBearGames:
Sometimes you might need to want direct control over the name that Godot sees for your method or property. For instance, maybe you want Godot/GDScript to see a different name for a property than Swift does:

@Export(explicitName = "my_param")
var myParamVariant: Variant? = nil

var myParam: String? { 
get { return String.fromVariant(myParamVariant)
}

That way the "myParam" property is nicely usable in Swift while maintaining a familiar name on the Godot side as well.

### DIFF
--- a/Sources/SwiftGodotRuntime/MacroDefs.swift
+++ b/Sources/SwiftGodotRuntime/MacroDefs.swift
@@ -32,9 +32,11 @@ public enum ClassBehavior: Int {
 ///
 /// The parameters and returns type of the function must be `_GodotBridgeable`.
 ///
-/// - Parameter autoSnakeCase: if `true` (default value is `false`), the function name will be automatically translated from `camelCase` to `snake_case` when exposed to Godot
+/// - Parameter autoSnakeCase: if `true` (default value is `false`), the function name will be automatically translated from `camelCase` to `snake_case` when exposed to Godot. `explicitName` must be set to nil to use autoSnakeCase.
+/// - Parameter explicitName: If not nil, this will be used as the function name exported to Godot exactly as written.  If using `explicitName`, `autoSnakeCase` must be false.
+///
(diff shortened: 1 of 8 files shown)


Sources/SwiftGodotRuntime/MacroDefs.swift
```diff
@@ -32,9 +32,11 @@ public enum ClassBehavior: Int {
 ///
 /// The parameters and returns type of the function must be `_GodotBridgeable`.
 ///
-/// - Parameter autoSnakeCase: if `true` (default value is `false`), the function name will be automatically translated from `camelCase` to `snake_case` when exposed to Godot
+/// - Parameter autoSnakeCase: if `true` (default value is `false`), the function name will be automatically translated from `camelCase` to `snake_case` when exposed to Godot. `explicitName` must be set to nil to use autoSnakeCase.
+/// - Parameter explicitName: If not nil, this will be used as the function name exported to Godot exactly as written.  If using `explicitName`, `autoSnakeCase` must be false.
+///
 @attached(peer, names: prefixed(_mproxy_), prefixed(_pproxy_))
-public macro Callable(autoSnakeCase: Bool = false) = #externalMacro(module: "SwiftGodotMacroLibrary", type: "GodotCallable")
+public macro Callable(autoSnakeCase: Bool = false, explicitName: String? = nil) = #externalMacro(module: "SwiftGodotMacroLibrary", type: "GodotCallable")
```
elijah-semyonov: Macro behave like functions declarations - you can define a different one with the same name and different parameters, so exclusive behaviors are not crammed into the same signature

ReddishBearGames: Addressed in subsequent commit

